PR #25, opened 2021-09-23
Add ARM64 support

stanek-michal:


Trinity2019 (approved):

contrib/elftoolchain/common/native-elf-format
```diff
@@ -40,7 +40,7 @@ $1 ~ "Machine:" {
         } else if (match($0, "PowerPC64")) {
             elfarch = "EM_PPC64";
         } else {
-            elfarch = "unknown";
+            elfarch = "EM_AARCH64";
```
dwhyrock, line 43: Can we match on something to determine `aarch64` instead of treating it as a default?  Seems like the default case could cause unknown issues down the line.

jrmolin (approved):
I guess those header files are standard. I saw the license blurb at the top. Seems fine to me. Hope it works!

contrib/elftoolchain/common/native-elf-format
dwhyrock: i just saw that this is in the `contrib/` dir.  How does modifying it in here work?  Do we have to publish changes, or what?

Do we need to log where we pulled these changes from?

dwhyrock (commented):
comment and question about contrib

dwhyrock (approved):

PR #27, opened 2021-10-13
Add option to allow IP subnets

stanek-michal:
The allowed_subnets map will contain user-defined IPs and subnets.
The allowed_IPs map will now only be used by the KprobeConnectHook program
to dynamically add destination IPs that allowed processes want to connect to.

The TcFilter network filter checks both maps and allows the packet to
be transmitted if it finds a match in either map.

### DIFF
--- a/GPL/HostIsolation/TcFilter/TcFilter.bpf.c
+++ b/GPL/HostIsolation/TcFilter/TcFilter.bpf.c
@@ -70,24 +70,24 @@ allow_destination_IP(
         .data = ip->daddr,
     };
 
-    /* Check allowed subnets first */
-    elem = bpf_map_lookup_elem(&allowed_subnets, &lpm_key);
+    /* Check allowed IPs map first */
+    elem = bpf_map_lookup_elem(&allowed_IPs, &ip->daddr);
     if (elem)
     {
-        /* IP matches allowed subnet */
+        /* IP is allowed */
         return 1;
     }
 
-    /* Now check allowed IPs map */
-    elem = bpf_map_lookup_elem(&allowed_IPs, &ip->daddr);
+    /* Now check allowed subnets */
+    elem = bpf_map_lookup_elem(&allowed_subnets, &lpm_key);
     if (!elem)
     {
-        /* destination IP not in allowlist - reject */
+        /* destination IP not within any allowed subnets - reject */
         return 0;
     }
     else
     {
-        /* IP is allowed */
+        /* IP matches allowed subnet */
         return 1;
     }
 }
--- a/GPL/HostIsolation/TcFilter/TcFilter.bpf.c
+++ b/GPL/HostIsolation/TcFilter/TcFilter.bpf.c
@@ -32,6 +32,14 @@
 // you took NULL for granted didn't you :)
 #define NULL 0
 
+#define BPF_F_NO_PREALLOC	(1U << 0)
+
+/* Key of an a BPF_MAP_TYPE_LPM_TRIE entry */
+struct bpf_lpm_trie_key {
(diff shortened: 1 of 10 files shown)


GPL/HostIsolation/TcFilter/TcFilter.bpf.c
```diff
@@ -32,6 +32,14 @@
 // you took NULL for granted didn't you :)
 #define NULL 0
 
+#define BPF_F_NO_PREALLOC	(1U << 0)
+
+/* Key of an a BPF_MAP_TYPE_LPM_TRIE entry */
+struct bpf_lpm_trie_key {
+	__u32	prefixlen;	/* up to 32 for AF_INET, 128 for AF_INET6 */
+	__u32	data;	/* Arbitrary size */
```
nicholasberlin, line 40: Perhaps this should be large enough to hold an ipv6 address.

non-GPL/HostIsolationMapsUtil/UpdateMaps.c
```diff
@@ -44,23 +66,64 @@ ebpf_map_allowed_pids_add(uint32_t pid)
...
 int
 ebpf_map_allowed_pids_clear()
 {
-    return ebpf_clear_map(EBPF_ALLOWED_PIDS_MAP_PATH);
+    return ebpf_clear_map(EBPF_ALLOWED_PIDS_MAP_PATH, EBPF_MAP_ALLOWED_PIDS);
 }
 
 static int
-ebpf_update_map(const char *map_path, uint32_t key, uint32_t val)
+ebpf_create_map(const char *map_path,
```
nicholasberlin, line 91: `map_path` is not used in this function. perhaps remove it?

nicholasberlin (approved):
Some clean up / simplification suggestions but LGTM

non-GPL/HostIsolationMapsUtil/UpdateMaps.c
stanek-michal: I'll remove it

non-GPL/HostIsolationMapsUtil/UpdateMaps.c
```diff
@@ -44,23 +65,63 @@ ebpf_map_allowed_pids_add(uint32_t pid)
...
+                             ebpf_maps[map_id].map_flags);
+
+    if (fd < 0)
+    {
+        ebpf_log("Error creating map\n");
+        rv = -1;
+        goto cleanup;
+    }
+
+    *map_fd = fd;
```
dwhyrock, line 117: minor nitpick about the lack of a NULL check on `map_fd`, although it appears this function is only used internally in this file w/ refs to local variables.  If you don't want to add a NULL check, perhaps add a comment at the top of the function stating why.

non-GPL/HostIsolationMapsUtil/UpdateMaps.h
```diff
@@ -6,9 +6,11 @@
  * you may not use this file except in compliance with the Elastic License 2.0.
  */
 
+#ifndef EBPF_UPDATE_MAPS_H
+#define EBPF_UPDATE_MAPS_H
```
dwhyrock, line 10: Using `#pragma once` could simplify this.

dwhyrock (approved):
A couple comments, but nothing to hold up this PR.

stanek-michal:
> A couple comments, but nothing to hold up this PR.

I'll make a note to fix those in the next PR if you don't mind.

fntlnz:
Good job @stanek-michal  I did a review to this and code looked good to me.
 I didn’t have a chance (had issues with my environment)  to run the prog test run tests in this repo and see if they cover the new use case.

I’ll test those and in case there are any tests not passing I’ll open an issue. I suspect not however.

PR #33, opened 2021-11-05
Fix map creation on AWS Linux

stanek-michal:
AWS has an old kernel (4.14) where struct bpf_map did not have
the name field. Creating a map with a name caused an error.
Instead, create a map with no name. The maps are pinned to BPF FS
so bpftool should still show the names properly with 'map list'.


nicholasberlin (approved):

fntlnz (approved):

leodido:
This is the reason I like eBPF.

2 ours debugging (in 3-4 persons) and then it resolves to one line change! :D

PR #35, opened 2021-11-08
tests: BPFTcFilterTests map unpin on teardown

fntlnz:
Since the maps are pinned to the filesystem
the tests might be flaky if the maps are not cleared at every iteration.

This test could be made more pure by non pinning the map. However we
really want to test that behavior as well because we rely on pinning and
non pinned maps could behave differently because they have a different
lifecycle.

### DIFF
--- a/GPL/HostIsolation/TcFilter/BPFTcFilterTests.cpp
+++ b/GPL/HostIsolation/TcFilter/BPFTcFilterTests.cpp
@@ -95,12 +95,20 @@ protected:
 
         m_prog_fd = bpf_program__fd(prog);
     }
+
     virtual void
     TearDown() override
     {
+        struct bpf_map *allowed_ips_map = bpf_object__find_map_by_name(m_obj, "allowed_IPs");
+        bpf_map__unpin(allowed_ips_map, bpf_map__get_pin_path(allowed_ips_map));
+
+        struct bpf_map *allowed_subnets_map = bpf_object__find_map_by_name(m_obj, "allowed_subnets");
+        bpf_map__unpin(allowed_subnets_map, bpf_map__get_pin_path(allowed_subnets_map));
+
         bpf_object__close(m_obj);
         m_prog_fd = -1;
     }
+
     static void
     SetUpTestSuite()
     {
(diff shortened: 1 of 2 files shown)


GPL/HostIsolation/TcFilter/BPFTcFilterTests.cpp
```diff
@@ -95,12 +95,20 @@ class BPFTcFilterTests : public ::testing::Test
 
         m_prog_fd = bpf_program__fd(prog);
     }
+
     virtual void
     TearDown() override
     {
+        struct bpf_map *allowed_ips_map = bpf_object__find_map_by_name(m_obj, "allowed_IPs");
```
fntlnz, line 102: Passing the map name as a string here is done on purpose. We do not want to rely on the userspace loading logic from `libeBPF` here.

GPL/HostIsolation/TcFilter/BPFTcFilterTests.cpp
```diff
@@ -95,12 +95,20 @@ class BPFTcFilterTests : public ::testing::Test
...
+        bpf_map__unpin(allowed_ips_map, bpf_map__get_pin_path(allowed_ips_map));
```
nicholasberlin, line 103: I'm not aware a situation where this could occur, but it looks generally possible for `bpf_object__find_map_by_name` to return `NULL`.

Should probably check for NULL before calling this and line 106.

nicholasberlin: I prefer a test failure, than a test crash. Easier for CI to gather logs, easier to see where things went wrong, don't need a debugger.

leodido (commented):
LGTM 🔥 

Left just a comment

GPL/HostIsolation/TcFilter/BPFTcFilterTests.cpp
fntlnz: I understand the desire, let me send a patch for that !

leodido (approved):
LGTM

stanek-michal (approved):
LGTM

nicholasberlin (approved):

PR #38, opened 2021-11-12
proposal: libebpf.h header for feature detection and events

fntlnz:
Hello everyone 👋 

Based on the discussions we had with many of the maintainers here I'm trying to create a starting point for us to work on.

**Important things to remember**

- This is opened early to nurture collaboration, might be all wrong, suggest, send patches, let's discuss!
- This is as mine as it is yours, let's look at it as a starting point for what it will be
- This does not imply any implementation details, this is how our black box will look like outside. Anything in this file is something we are OK in having in a public API that will keep bc.
- Anything that is not to be kept public does not go into this file (like any implementation detail)
- I created a new `include/` folder, probably the wrong place but a good one for it to be easy to find while we work on it now. - This is not in CMake or whatever and probably does not even compile.
-We will need to make this work for the existing HostIsolation code too but I decided to start from scratch with this header file so we can focus on the design first
- I didn't include the loading and all the other boilerplate things for BPF coming from libbpf (or skeletons) - we can think on where we want to put those
- We will probably have specific APIs for specific features (like HostIsolation, Processes or File Events), we can think here on how to add those too, I didn't do it already because I'm thinking about the foundations first.

### DIFF
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -71,7 +71,7 @@ ebpf_cap_set ebpf_event__detect_capabilities(
 /* Combination of events user is interested in. Mapping of events
  * to probe that provides that event is stored internally. User
  * does not need to know which probes to load */
-typedef uint32_t ebpf_event_set;
+typedef uint64_t ebpf_event_set;
 
 /* allocates a new context based on requested events and capabilities.
  *
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -38,6 +38,10 @@ typedef enum ebpf_event_type {
     EVENT_NETWORK_IPV6_CONNECTION_ACCEPTED  = 9,
     EVENT_NETWORK_IPV6_RECONNECT_ATTEMPTED  = 10,
     EVENT_NETWORK_IPV6_HTTP_REQUEST         = 11,
+    EVENT_PROCESS_FORK,
+    EVENT_PROCESS_EXEC,
+    EVENT_PROCESS_EXIT,
+    EVENT_PROCESS_TTY_OUTPUT,
     EBPF_EVENT_MAX
 } ebpf_event_type;
 
@@ -47,9 +51,56 @@ struct ebpf_event {
     void *data;
 };
 
+/* opaque context, will contain union with struct perf_buffer,
+ * struct ring_buffer,
+ *
+ * we need a way to map from capability/event sets to probe sets,
+ * and we could store the cap/event sets directly or cache the result
+ * of the mapping in the context */
+typedef struct ebpf_event_ctx;
+
 typedef int (*event_handler_fn)(struct ebpf_event *event, size_t);
 
-void *ebpf_event__buffer_new(event_handler_fn handler);
-void ebpf_event__next(void *buffer);
+typedef uint32_t ebpf_cap_set;
+
+/* result will be ((ACTUAL AND INTERESTED) OR FORCED) */
+ebpf_cap_set ebpf_event__detect_capabilities(
+        ebpf_cap_set interested,
+        ebpf_cap_set forced)
+
+/* Combination of events user is interested in. Mapping of events
+ * to probe that provides that event is stored internally. User
+ * does not need to know which probes to load */
+typedef uint32_t ebpf_event_set;
+
+/* allocates a new context based on requested events and capabilities.
+ *
+ * returns a positive int that represents an fd, which
+ * can be used with epoll on success. returns -1 and
+ * sets errno on failure
+ */
+int ebpf_event__buffer_new(
+        ebpf_event_ctx **ctx,
+        ebpf_cap_set,
+        ebpf_event_set);
+
+/* this func will not return data unless fd is ready
+ * so user will supply their own epoll and call this
+ * func when fd is readable. I think passing a handler
+ * function is more versatile than passing a buffer?
+ * no strong feelings on that. The handler will likely
+ * dispatch one of another layer of callbacks switched
+ * on event type.
+ *
+ * We should also discuss blocking/nonblocking semantics
+ * here. I think this function could return -1 and set
+ * errno to -EAGAIN if the context is in a nonblocking
+ * state but no data is available. */
+int ebpf_event__next(
+        ebpf_event_ctx *ctx,
+        event_handler_fn *handler_func);
+
+void ebpf_event__cleanup(
+        ebpf_event_ctx *ctx);
 
 #endif // LIBEBPF_H_
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -59,7 +59,7 @@ struct ebpf_event {
  * of the mapping in the context */
 typedef struct ebpf_event_ctx;
 
-typedef int (*event_handler_fn)(struct ebpf_event *event, size_t);
+typedef int (*ebf_event_handler_fn)(struct ebpf_event *event, size_t);
 
 typedef uint64_t ebpf_cap_set;
 
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -61,7 +61,7 @@ typedef struct ebpf_event_ctx;
 
 typedef int (*event_handler_fn)(struct ebpf_event *event, size_t);
 
-typedef uint32_t ebpf_cap_set;
+typedef uint64_t ebpf_cap_set;
 
 /* result will be ((ACTUAL AND INTERESTED) OR FORCED) */
 ebpf_cap_set ebpf_event__detect_capabilities(
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: LicenseRef-Elastic-License-2.0
+
+/*
+ * Copyright 2021 Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+ * or more contributor license agreements. Licensed under the Elastic License 2.0;
+ * you may not use this file except in compliance with the Elastic License 2.0.
+ */
+
+// This file is just a spec for discussion, not included in builds or anything.
+// This does not compile and it's probably not in the folder we want it in.
+
+#ifndef LIBEBPF_H_
+#define LIBEBPF_H_
+
+#include <stdint.h>
+#include <sys/types.h>
+
+/* Kernel Features */
+enum ebpf_kernel_features {
+    EBPF_KERNEL_CAPABLE_BPF = (1 << 0),
+    EBPF_KERNEL_CAPABLE_RINGBUF = (1 << 1),
+    EBPF_KERNEL_CAPABLE_BTF = (1 << 2),
+};
+
+int ebpf_kernel__detect_features();
+
+/* Events fetching and manipulation */
+typedef enum ebpf_event_type {
+    EBPF_EVENT_TYPE_UNSPEC = 0,
+    EBPF_EVENT_TYPE_NETWORK_UPDATE = 1,
+    EVENT_NETWORK_IPV4_CONNECTION_ATTEMPTED = 2,
+    EVENT_NETWORK_IPV4_DISCONNECT_RECEIVED  = 3,
+    EVENT_NETWORK_IPV4_CONNECTION_ACCEPTED  = 4,
+    EVENT_NETWORK_IPV4_RECONNECT_ATTEMPTED  = 5,
+    EVENT_NETWORK_IPV4_HTTP_REQUEST         = 6,
+    EVENT_NETWORK_IPV6_CONNECTION_ATTEMPTED = 7,
+    EVENT_NETWORK_IPV6_DISCONNECT_RECEIVED  = 8,
+    EVENT_NETWORK_IPV6_CONNECTION_ACCEPTED  = 9,
+    EVENT_NETWORK_IPV6_RECONNECT_ATTEMPTED  = 10,
+    EVENT_NETWORK_IPV6_HTTP_REQUEST         = 11,
+    EBPF_EVENT_MAX
+} ebpf_event_type;
+
+struct ebpf_event {
+    uint64_t ts;
+    ebpf_event_type type;
+    void *data;
+};
+
+typedef int (*event_handler_fn)(struct ebpf_event *event, size_t);
+
+void *ebpf_event__buffer_new(event_handler_fn handler);
+void ebpf_event__next(void *buffer);
+
+#endif // LIBEBPF_H_
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -59,7 +59,7 @@ struct ebpf_event {
  * of the mapping in the context */
 typedef struct ebpf_event_ctx;
 
-typedef int (*ebf_event_handler_fn)(struct ebpf_event *event, size_t);
+typedef int (*ebpf_event_handler_fn)(struct ebpf_event *event, size_t);
 
 typedef uint64_t ebpf_cap_set;
 
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -44,7 +44,7 @@ enum ebpf_event_type {
 struct ebpf_event {
     uint64_t ts;
     uint64_t type;
-    void *data;
+    char data[];
 };
 
 /* opaque context, will contain union with struct perf_buffer,
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -53,7 +53,7 @@ struct ebpf_event {
  * we need a way to map from capability/event sets to probe sets,
  * and we could store the cap/event sets directly or cache the result
  * of the mapping in the context */
-typedef struct ebpf_event_ctx;
+struct ebpf_event_ctx;
 
 typedef int (*ebpf_event_handler_fn)(struct ebpf_event *event, size_t);
 
@@ -76,7 +76,7 @@ typedef uint64_t ebpf_event_set;
  * sets errno on failure
  */
 int ebpf_event__buffer_new(
-        ebpf_event_ctx **ctx,
+        struct ebpf_event_ctx **ctx,
         ebpf_cap_set,
         ebpf_event_set);
 
@@ -93,10 +93,10 @@ int ebpf_event__buffer_new(
  * errno to -EAGAIN if the context is in a nonblocking
  * state but no data is available. */
 int ebpf_event__next(
-        ebpf_event_ctx *ctx,
-        event_handler_fn *handler_func);
+        struct ebpf_event_ctx *ctx,
+        ebpf_event_handler_fn *handler_func);
 
 void ebpf_event__cleanup(
-        ebpf_event_ctx *ctx);
+        struct ebpf_event_ctx *ctx);
 
 #endif // LIBEBPF_H_
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -12,8 +12,10 @@
 #ifndef LIBEBPF_H_
 #define LIBEBPF_H_
 
+#if !defined(__KERNEL__) && !defined(__VMLINUX__)
 #include <stdint.h>
 #include <sys/types.h>
+#endif // !defined(__KERNEL__) && !defined(__VMLINUX__)
 
 /* Kernel Features */
 enum ebpf_kernel_features {
@@ -21,8 +23,6 @@ enum ebpf_kernel_features {
     EBPF_KERNEL_CAPABLE_RINGBUF = (1 << 1),
     EBPF_KERNEL_CAPABLE_BTF = (1 << 2),
 };
-
-int ebpf_kernel__detect_features();
 
 /* Events fetching and manipulation */
 enum ebpf_event_type {
@@ -46,6 +46,10 @@ struct ebpf_event {
     uint64_t type;
     char data[];
 };
+
+#ifndef __KERNEL__
+
+int ebpf_kernel__detect_features();
 
 /* opaque context, will contain union with struct perf_buffer,
  * struct ring_buffer,
@@ -99,4 +103,5 @@ int ebpf_event__next(
 void ebpf_event__cleanup(
         struct ebpf_event_ctx *ctx);
 
+#endif // ifndef __KERNEL__
 #endif // LIBEBPF_H_
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -49,8 +49,6 @@ struct ebpf_event {
 
 #ifndef __KERNEL__
 
-int ebpf_kernel__detect_features();
-
 /* opaque context, will contain union with struct perf_buffer,
  * struct ring_buffer,
  *
@@ -63,10 +61,8 @@ typedef int (*ebpf_event_handler_fn)(struct ebpf_event *event, size_t);
 
 typedef uint64_t ebpf_cap_set;
 
-/* result will be ((ACTUAL AND INTERESTED) OR FORCED) */
-ebpf_cap_set ebpf_event__detect_capabilities(
-        ebpf_cap_set interested,
-        ebpf_cap_set forced);
+/* result will be ACTUAL | FORCED) */
+ebpf_cap_set ebpf_event__detect_capabilities(ebpf_cap_set forced);
 
 /* Combination of events user is interested in. Mapping of events
  * to probe that provides that event is stored internally. User
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -26,23 +26,19 @@ int ebpf_kernel__detect_features();
 
 /* Events fetching and manipulation */
 typedef enum ebpf_event_type {
-    EBPF_EVENT_TYPE_UNSPEC = 0,
-    EBPF_EVENT_TYPE_NETWORK_UPDATE = 1,
-    EVENT_NETWORK_IPV4_CONNECTION_ATTEMPTED = 2,
-    EVENT_NETWORK_IPV4_DISCONNECT_RECEIVED  = 3,
-    EVENT_NETWORK_IPV4_CONNECTION_ACCEPTED  = 4,
-    EVENT_NETWORK_IPV4_RECONNECT_ATTEMPTED  = 5,
-    EVENT_NETWORK_IPV4_HTTP_REQUEST         = 6,
-    EVENT_NETWORK_IPV6_CONNECTION_ATTEMPTED = 7,
-    EVENT_NETWORK_IPV6_DISCONNECT_RECEIVED  = 8,
-    EVENT_NETWORK_IPV6_CONNECTION_ACCEPTED  = 9,
-    EVENT_NETWORK_IPV6_RECONNECT_ATTEMPTED  = 10,
-    EVENT_NETWORK_IPV6_HTTP_REQUEST         = 11,
-    EVENT_PROCESS_FORK,
-    EVENT_PROCESS_EXEC,
-    EVENT_PROCESS_EXIT,
-    EVENT_PROCESS_TTY_OUTPUT,
-    EBPF_EVENT_MAX
+    EBPF_EVENT_PROCESS_FORK = (1 << 1),
+    EBPF_EVENT_PROCESS_EXEC = (1 << 2),
+    EBPF_EVENT_PROCESS_EXIT = (1 << 3),
+    EBPF_EVENT_PROCESS_TTY_OUTPUT = (1 << 4),
+    EBPF_EVENT_FILE_CREATE = (1 << 5),
+    EBPF_EVENT_FILE_DELETE = (1 << 6),
+    EBPF_EVENT_FILE_RENAME  = (1 << 7),
+    EBPF_EVENT_FILE_OPEN = (1 << 8),
+    EBPF_EVENT_FILE_EXCHANGE = (1 << 9),
+    EBPF_EVENT_INTERNAL_FILE_OPEN = (1 << 10),
+    EBPF_EVENT_INTERNAL_FILE_CLOSE = (1 << 11),
+    EBPF_EVENT_INTERNAL_FILE_CHDIR = (1 << 12),
+    EBPF_EVENT_INTERNAL_FILE_DUP = (1 << 13),
 } ebpf_event_type;
 
 struct ebpf_event {
@@ -61,12 +57,12 @@ typedef struct ebpf_event_ctx;
 
 typedef int (*ebpf_event_handler_fn)(struct ebpf_event *event, size_t);
 
-typedef uint64_t ebpf_cap_set;
+typedef ebpf_event_type ebpf_cap_set;
 
 /* result will be ((ACTUAL AND INTERESTED) OR FORCED) */
 ebpf_cap_set ebpf_event__detect_capabilities(
         ebpf_cap_set interested,
-        ebpf_cap_set forced)
+        ebpf_cap_set forced);
 
 /* Combination of events user is interested in. Mapping of events
  * to probe that provides that event is stored internally. User
--- a/include/libebpf.h
+++ b/include/libebpf.h
@@ -25,7 +25,7 @@ enum ebpf_kernel_features {
 int ebpf_kernel__detect_features();
 
 /* Events fetching and manipulation */
-typedef enum ebpf_event_type {
+enum ebpf_event_type {
     EBPF_EVENT_PROCESS_FORK = (1 << 1),
     EBPF_EVENT_PROCESS_EXEC = (1 << 2),
     EBPF_EVENT_PROCESS_EXIT = (1 << 3),
@@ -39,11 +39,11 @@ typedef enum ebpf_event_type {
     EBPF_EVENT_INTERNAL_FILE_CLOSE = (1 << 11),
     EBPF_EVENT_INTERNAL_FILE_CHDIR = (1 << 12),
     EBPF_EVENT_INTERNAL_FILE_DUP = (1 << 13),
-} ebpf_event_type;
+};
 
 struct ebpf_event {
     uint64_t ts;
-    ebpf_event_type type;
+    uint64_t type;
     void *data;
 };
 
@@ -57,7 +57,7 @@ typedef struct ebpf_event_ctx;
 
 typedef int (*ebpf_event_handler_fn)(struct ebpf_event *event, size_t);
 
-typedef ebpf_event_type ebpf_cap_set;
+typedef uint64_t ebpf_cap_set;
 
 /* result will be ((ACTUAL AND INTERESTED) OR FORCED) */
 ebpf_cap_set ebpf_event__detect_capabilities(


fntlnz:
cc @ferullo @tabell @leodido @nicholasberlin @mattnite @norrietaylor @stanek-michal

m-sample:
Thanks for getting started on this - I spent part of the day looking at how we would unify a perf_buffer xor ring_buffer api and have some text I will share shortly.

In this header file, how would we discriminate an HTTP event from a regular TCP event?  Usually TLS is done in user space so content inspection would be uprobe-y over openssl, wolfssl, etc & version pain-y. Or am I missing something? I know about KTLS but I didn't think it was pervasive.

I'd recommend connection events include the protocol too and have protocol specific content - e.g. TCP vs SCTP etc.  Datagram protocols might require a msg_send_x_proto event but that would be really chatty so a BPF Map with summary stats might be wiser.  However,  I'd suggest we defer network event design for now now -hand have an initial focus on process fork/exec/exit, tty input/output and new sessions.

include/libebpf.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include <stdint.h>
+#include <sys/types.h>
+
+/* Kernel Features */
+enum ebpf_kernel_features {
+    EBPF_KERNEL_CAPABLE_BPF = (1 << 0),
+    EBPF_KERNEL_CAPABLE_RINGBUF = (1 << 1),
+    EBPF_KERNEL_CAPABLE_BTF = (1 << 2),
+};
```
ferullo, line 23: This API limits us to 64 capabilities. I bet that's fine, just calling it out for thought.

include/libebpf.h
```diff
@@ -0,0 +1,56 @@
...
+
+ebpf_kernel_features
+ebpf_kernel__detect_features();
```
ferullo, line 26: Somehow I think the caller should be able to override internal capability detection. If you expect that this library would maintain any state about what capabilities it can handle then this function would benefit from an input param where capabilities can be forced on or off. However, if you expect that the caller would pass these capabilities back into this library in later API calls then this API would work.  Am I making sense?

include/libebpf.h
```diff
@@ -0,0 +1,56 @@
...
+
+/* Events fetching and manipulation */
+enum ebpf_event_type {
```
ferullo, line 29: I've gone back and forth on this enum. It seems to unify all possible probes this library will implement. There would be no concept of probe "namespacing". I can't think of a drawback to that approach and it certainly does simplify things. Even if not all probes were available or loaded at run time, I can't think of how a universal list of what all probes _could_ produce would be a problem.

So, I think this enum makes sense, with I think an implied understanding that the actual values should be scrapped in a real implementation and replaced with values that make sense for the how the probes are actually written.

include/libebpf.h
```diff
@@ -0,0 +1,56 @@
...
+    EVENT_NETWORK_IPV6_CONNECTION_ACCEPTED  = 9,
+    EVENT_NETWORK_IPV6_RECONNECT_ATTEMPTED  = 10,
+    EVENT_NETWORK_IPV6_HTTP_REQUEST         = 11,
+    EBPF_EVENT_MAX
+};
+
+struct ebpf_event {
+    uint64_t ts;
+    ebpf_event_type type;
+    void *data;
```
ferullo, line 48: why not `char data[0]` so only one allocation is needed? also, why no size value. even for a known event type, i expect there could be a variable sized data buffer.

include/libebpf.h
```diff
@@ -0,0 +1,56 @@
...
+};
+
+typedef int (*event_handler_fn)(ebpf_event *event, size_t size);
+
+void *ebpf_event__buffer_new(event_handler_fn handler);
+void ebpf_event__next(void *buffer);
```
ferullo, line 54: can you describe/demo the API these 4 lines would implement?

ferullo (commented):
Thanks for starting this discussion!

include/libebpf.h
fntlnz: You are right, I did put this in the callback but it does not make sense since the first type in the callback is known. I'll move it here.

fntlnz: Updated

fntlnz: `char data[0]` and `void *` do not have any differences in usage here as far as I can observe in my systems.
I picked `void *` to explicitly state that this type is to be cast to whatever type is needed for the specific event we are dealing with in the callback.

I'm preparing some examples to make it easier to explain.

include/libebpf.h
fntlnz: That's exactly what I was thinking here.

Of course the ones I did put here are just examples.

include/libebpf.h
fntlnz: Yeah that is a limitation of this approach. I thought about that problem but the advantage of being able to express the capabilities with a single value really made it worth for the kind of usage we need to make of this.

I believe that we can shrink multiple related capabilities in the same constant if we think this will become a problem.

include/libebpf.h
fntlnz: I've been thinking about this as an API that is both internal and public.

The implementations exposed by library, e.g: `ebpf_event__buffer_new` will need to use it **internally** to decide what to do. I've been thinking about making this also public because in this way the caller has the ability to decide autonomously.

This will be a constant scalar value compared with bitwise operations and the way I see it is that it is computed only the first time this function is called, internally or externally and then stored.

If we want to make it editable to force capabilities off we can have an `ebpf_kernel__set_features(ebpf_kernel_features)` to override the bitmask. The ovverride will need to be a subset of the detected value.

fntlnz:
@m-sample 

>  I spent part of the day looking at how we would unify a perf_buffer xor ring_buffer api and have some text I will share shortly.

Thanks! Looking forward it!


> In this header file, how would we discriminate an HTTP event from a regular TCP event? Usually TLS is done in user space so content inspection would be uprobe-y over openssl, wolfssl, etc & version pain-y. Or am I missing something? I know about KTLS but I didn't think it was pervasive.

> I'd recommend connection events include the protocol too and have protocol specific content - e.g. TCP vs SCTP etc. Datagram protocols might require a msg_send_x_proto event but that would be really chatty so a BPF Map with summary stats might be wiser. However, I'd suggest we defer network event design for now - and have an initial focus on process fork/exec/exit, tty input/output and new sessions.


I think I might have confused you into thinking this header is about network events. The event types I proposed are just there to fill the enum with examples, we are really discussing the foundations of the API here  instead of the specific implementation for networking, files or processes. That can come in subsequent PRs once we agree on the foundations first.

include/libebpf.h
fntlnz: While working on the example implementation I had to switch len back to be a function parameterto  the handler callback. The reason for it is that the size is never inside the struct from the outside when the data is passed down.

intxgo: so `*data` points to another memory, I thought that it was meant to be the memory adjacent to the struct.

fntlnz: yes, this is a pointer to the memory in the actual buffer implementation (ring buffer or perf buffer). 

The implementation for ring buffer can be seen here https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/tree/tools/lib/bpf/ringbuf.c#n231?id=8bc117bcc7bcd2eb1839d34f8648aeae9845a49b

include/libebpf.h
leodido: I think it's worth linking #39 here for discussion on this (and only this) topic (feature detection for eBPF) :)

include/libebpf.h
fntlnz: The usage of this part of the API would look something like this

```c
#include <libebpf.h>
#include <stdio.h>

int event_handler(struct ebpf_event *event, size_t size)
{
    printf("event received, type: %d", (int)event->type);
    return 0;
}

int main(int argc, char const *argv[])
{
    void *buf = ebpf_event__buffer_new(event_handler);
    if (buf == NULL) {
        printf("failed to allocate buffer\n");
        return 1;
    }
    ebpf_event__next(buf);
    return 0;
}
```

Where `event_handler` knows the type of the event and can then cast the `void *` inside of it to the actual struct based on the type itself. The type can also be used for mapping the type of this library to the type of the application that uses the library itself.

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+
+/* opaque context, will contain union with struct perf_buffer,
+ * struct ring_buffer,
+ *
+ * we need a way to map from capability/event sets to probe sets,
+ * and we could store the cap/event sets directly or cache the result
+ * of the mapping in the context */
+typedef struct ebpf_event_ctx;
+
+typedef int (*event_handler_fn)(struct ebpf_event *event, size_t);
```
nicholasberlin, line 62: What do the return values mean?

nicholasberlin:
```suggestion
typedef int (*event_handler_fn)(void *context, struct ebpf_event *event, size_t);
```
The function pointer implementation will likely need some context.

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+ * dispatch one of another layer of callbacks switched
+ * on event type.
+ *
+ * We should also discuss blocking/nonblocking semantics
+ * here. I think this function could return -1 and set
+ * errno to -EAGAIN if the context is in a nonblocking
+ * state but no data is available. */
+int ebpf_event__next(
+        ebpf_event_ctx *ctx,
+        event_handler_fn *handler_func);
```
nicholasberlin, line 101:
```suggestion
int ebpf_event__next(
        ebpf_event_ctx *ctx,
        event_handler_fn *handler_func,
        void *handler_context);
```

nicholasberlin: The `handler_context` would be blindly passed to `handler_func` by the implementation.

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+ * struct ring_buffer,
+ *
+ * we need a way to map from capability/event sets to probe sets,
+ * and we could store the cap/event sets directly or cache the result
+ * of the mapping in the context */
+typedef struct ebpf_event_ctx;
+
+typedef int (*event_handler_fn)(struct ebpf_event *event, size_t);
+
+typedef uint32_t ebpf_cap_set;
```
nicholasberlin, line 64:
```suggestion
typedef uint64_t ebpf_cap_set;
```

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+
+/* result will be ((ACTUAL AND INTERESTED) OR FORCED) */
+ebpf_cap_set ebpf_event__detect_capabilities(
+        ebpf_cap_set interested,
+        ebpf_cap_set forced)
+
+/* Combination of events user is interested in. Mapping of events
+ * to probe that provides that event is stored internally. User
+ * does not need to know which probes to load */
+typedef uint32_t ebpf_event_set;
```
nicholasberlin, line 74:
```suggestion
typedef uint64_t ebpf_event_set;
```

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+
+/* allocates a new context based on requested events and capabilities.
+ *
+ * returns a positive int that represents an fd, which
```
nicholasberlin, line 78: This will be problematic if multiple fds are desired. Perhaps, fds should be part of `ebpf_event_ctx`?

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+ * We should also discuss blocking/nonblocking semantics
+ * here. I think this function could return -1 and set
+ * errno to -EAGAIN if the context is in a nonblocking
+ * state but no data is available. */
+int ebpf_event__next(
+        ebpf_event_ctx *ctx,
+        event_handler_fn *handler_func);
+
+void ebpf_event__cleanup(
+        ebpf_event_ctx *ctx);
```
nicholasberlin, line 104: When will file descriptors be closed?

include/libebpf.h
tabell: I think this makes sense. I might call it something more generic eg. handler_data

include/libebpf.h
tabell: I assumed the cleanup function would close all fds. I can't think of a case where fd(s) should persist. If you agree I will add a comment to this effect.

include/libebpf.h
nicholasberlin: ok. updated this and the function pointer definition to be the same.

include/libebpf.h
tabell: This makes sense, however I think the name _context_ could be confusing here, as it is similar to ebpf_event_ctx. I would propose user_data or handler_data

include/libebpf.h
tabell: Yes you're correct. My thought was that the `ebpf_event_ctx` would contain a `struct perf_buffer` or `struct ring_buffer`, which would mean the fds would be accessible from there as well. There is no reason the `ebpf_event_ctx` could not store an array of perf_buffers (or even ring_buffers, if so desired). That would mean `ebpf_event__new_buffer` would simply return 0 on success, and a new function would be required to retrieve the fd(s) from an initialized `ebpf_event_ctx`

include/libebpf.h
nicholasberlin: updated to be: `handler_data`

include/libebpf.h
nicholasberlin: sounds good to me

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+    EVENT_NETWORK_IPV4_DISCONNECT_RECEIVED  = 3,
+    EVENT_NETWORK_IPV4_CONNECTION_ACCEPTED  = 4,
+    EVENT_NETWORK_IPV4_RECONNECT_ATTEMPTED  = 5,
+    EVENT_NETWORK_IPV4_HTTP_REQUEST         = 6,
+    EVENT_NETWORK_IPV6_CONNECTION_ATTEMPTED = 7,
+    EVENT_NETWORK_IPV6_DISCONNECT_RECEIVED  = 8,
+    EVENT_NETWORK_IPV6_CONNECTION_ACCEPTED  = 9,
+    EVENT_NETWORK_IPV6_RECONNECT_ATTEMPTED  = 10,
+    EVENT_NETWORK_IPV6_HTTP_REQUEST         = 11,
+    EVENT_PROCESS_FORK,
```
fntlnz, line 41: I know these are jus examples, however we want to be explicit with the indexes here even if the enum does that for us

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+ * and we could store the cap/event sets directly or cache the result
+ * of the mapping in the context */
+typedef struct ebpf_event_ctx;
+
+typedef int (*event_handler_fn)(struct ebpf_event *event, size_t);
+
+typedef uint64_t ebpf_cap_set;
+
+/* result will be ((ACTUAL AND INTERESTED) OR FORCED) */
+ebpf_cap_set ebpf_event__detect_capabilities(
```
fntlnz, line 67: An `ebpf_cap_set` is a bitmask of `ebpf_event_type` here? 

`ebpf_event__detect_capabilities` and `ebpf_kernel__detect_features` seem to be doing the same thing here.

I used `features` and not `capabilities` because capabilities in Linux are a well defined name that will create confusion if used for another thing. See `man 7 capabilities`.  Features on the other hand seem to be a very good name to me. BPF is a feature, BTF is a feature, Ring buffer is a feature.

While if we talk about capabilities, being able to administer the network `CAP_SYS_ADMIN` is a capability and so on..

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+        ebpf_cap_set interested,
+        ebpf_cap_set forced)
+
+/* Combination of events user is interested in. Mapping of events
+ * to probe that provides that event is stored internally. User
+ * does not need to know which probes to load */
+typedef uint64_t ebpf_event_set;
```
fntlnz, line 74: Is this intended to be a bitmask? 

`ebpf_event_type` is not in powers of two. It would also be impractical in this case since we really want a list of event types and not something to compare against.

You probably wanted something like

```suggestion
typedef uint64_t ebpf_event_set[EBPF_EVENT_MAX];
```

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+ * function is more versatile than passing a buffer?
+ * no strong feelings on that. The handler will likely
+ * dispatch one of another layer of callbacks switched
+ * on event type.
+ *
+ * We should also discuss blocking/nonblocking semantics
+ * here. I think this function could return -1 and set
+ * errno to -EAGAIN if the context is in a nonblocking
+ * state but no data is available. */
+int ebpf_event__next(
```
fntlnz, line 99: So this is the function that should consume the epoll fd?
How does this know the fd itself?

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+
+/* allocates a new context based on requested events and capabilities.
+ *
+ * returns a positive int that represents an fd, which
+ * can be used with epoll on success. returns -1 and
+ * sets errno on failure
+ */
+int ebpf_event__buffer_new(
+        ebpf_event_ctx **ctx,
+        ebpf_cap_set,
```
fntlnz, line 84: I'm confused on why the buffer allocation should know about the features here.
I understadn the desire but this could do it internally by calling the API for the features themselves, am I missing something?

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * We should also discuss blocking/nonblocking semantics
+ * here. I think this function could return -1 and set
+ * errno to -EAGAIN if the context is in a nonblocking
+ * state but no data is available. */
+int ebpf_event__next(
+        ebpf_event_ctx *ctx,
+        event_handler_fn *handler_func);
+
+void ebpf_event__cleanup(
```
fntlnz, line 103:
```suggestion
void ebpf_event_buffer__free(
```

`ebpf_event_buffer__free` is a more idomatic name. Wdyt?

include/libebpf.h
fntlnz, line 99: We are still talking about the event buffer here, not about the event, this should be
```suggestion
int ebpf_event_buffer__next(
```

include/libebpf.h
fntlnz: I also don't understand why this is on `event` now. The set of features is a thing about the kernel, not about the specific event.

include/libebpf.h
fntlnz: I don't really understand this, can you make examples?
It seems confusing to me that we have the fd as a return argument and as context

include/libebpf.h
nicholasberlin: Perhaps the event types *should* be powers of two. We want to pass many of them at the same time, they are effectively a bitmask.

I like a bitmask because the arguments in other functions are then `uint64_t`s. However, that limits us to 64 events types without having to change the API. Perhaps that's enough for a long time.

Alternatively, as you suggest, we could use an array of `uint8_t`s with  `EBPF_EVENT_MAX` set to something like 1024. That would provide more runway, but then the APIs would have pointers for parameters. Array parameters (pointers) are easy for library users to get wrong and hard for library authors to validate.

include/libebpf.h
nicholasberlin: Also, if we store fds in the `ebpf_event_ctx`, the structure will no longer be opaque to users.

Perhaps, we should add a CPU num argument, and enforce that this function returns a single event source.
Hmm, that means the library user will need to know if they are getting a single cpu's per buffer or system-wide ring buffer.

include/libebpf.h
nicholasberlin: I was wondering that too.

We need to pass the fd as a parameter, or assume that an `ebf_event_ctx` only represents a single fd.

include/libebpf.h
tabell: @fntlnz my original thought was to return a single fd as return value but now I think that will not scale to multiple fds. 

I think that we need the fds to remain within the context, but we also need a way for the user to obtain them, therefore I would propose a new function that receives a context and a pointer to some type of fd array, which it then populates.

eg. `void ebpf_event__retrieve_fds(ebpf_event_ctx *ctx, int *fds, size_t *fd_count)`

Maybe there are other suggestions for a way to return this data to the user

In any case, I would keep the return value as int for `ebpf_event__buffer_new` to indicate success (0) or error (-1) with errno set in the latter case.

include/libebpf.h
tabell: @fntlnz I think there are two separate sources of confusion here that I will try to explain.  
1. are capabilities the same as features? yes. I called it capabilities but I should have called it features. This will prevent confusion with linux system capabilities such as CAP_SYS_ADMIN. 
2. are features the same as events? no. the idea here is that the selection of which probes to load will require knowledge of both features and events. 

as an example of 2, imagine if we eventually have four probes. one for network events using perf buffers, one for process events using perf buffers, one for network events using ring buffers, and one for process events using ring buffers.

`ebpf_event__buffer_new` would consider the events requested and the features available and from that it could determine which probes to load and which type of buffer to create

fntlnz: Ok so if I understood correctly: 

- `s/capabilities/features/` is ok for you
- you want to pass the features to `ebpf_event__buffer_new` instead of having it figuring out internally

Is that all here?

include/libebpf.h
fntlnz: Ok I like returning a context, I was confused by the return of `ebpf_event__buffer_new` but if we use it  only for error reporting and not for fds I'm fine with this approach.

include/libebpf.h
fntlnz: The advantages of having a bitmask are:

- efficient bitwise comparisons
- single value to represent a container of many states
- Very low footprint memory
 
Disadvantages

- Limited to 64 elements


Advantages of an array
- Can represent more than 64 elements

Disadvantages of an array
- More difficult to use in the API for the final user (they have to fill an array)


I honestly would go for a bitmask as well, question is: 64 elements are enough? And if it is, how much room do we have for new events?

include/libebpf.h
```diff
@@ -0,0 +1,106 @@
...
+typedef uint64_t ebpf_event_set;
+
+/* allocates a new context based on requested events and capabilities.
+ *
+ * returns a positive int that represents an fd, which
+ * can be used with epoll on success. returns -1 and
+ * sets errno on failure
+ */
+int ebpf_event__buffer_new(
+        ebpf_event_ctx **ctx,
```
fntlnz, line 83:
```suggestion
        ebpf_event_ctx *ctx,
```

Should be this?

include/libebpf.h
fntlnz: Apologies @tabell - I called them capabilities all over the place and asked you to not do it 😓

include/libebpf.h
mattnite: If we end up wanting to keep the interface light we could do something like the following:
```c
struct ebpf_context;

// an abstraction over a perf buffer or a ring buffer, calling it a stream is
// an attempt at finding a more generic term
struct ebpf_event_stream;

// placeholder
struct ebpf_event;

// gets a pointer and a length to an array of opaque pointers to event streams
// for an ebpf_context. In the ringbuffer case we'd get a length of one, and
// for perfbuffers we'd see one or more. The user could then manage them with
// whatever event loop setup they want
ssize_t ebpf_get_event_streams(
    struct ebpf_context *ctx,
    struct ebpf_event_stream **out);

int ebpf_event_stream__get_fd(
        struct ebpf_event_stream *stream);

int ebpf_event_stream__read(
        struct ebp_event_stream *stream,
        struct ebpf_event *out);
```

include/libebpf.h
nicholasberlin: A single pointer implies the user of the library knows at least the size of an `ebpf_event_ctx`. I think double pointer makes sense here if we want to keep the type opaque.

include/libebpf.h
fntlnz: @mattnite  thanks, this makes more sense, please update the header to include this so we can evolve from that.

nicholasberlin: I think we should add another API to consume events in a system wide order.

```
// Setup however many `ebpf_event_stream`s are needed,
// poll them, order the resulting events, and feed them
// to the consumer until consumer returns 1.
// After which, fds are closed, buffers are freed, etc, all internal to this function.
int ebpf_consume_ordered_event_stream(
    struct ebpf_context *ctx,
    ebf_event_handler_fn *consumer);
```

fntlnz: Would the different file descriptors guarantee ordering in this case?

nicholasberlin: For the case of perf_event_buffers, no. This function would have to handle that.

fntlnz: I'm starting thinking about the fact that this API should retunr only one fd like it was before and all the complexity should be hidden  to the caller.

So, in case this is implemented with a ringbuf we just use that fd, in case of perf buffer we would have to order the events in a new fd that we return.

Wdyt?

include/libebpf.h
tabell: The primary reason to expose this is to allow the user of the library more control and visibility, eg. for the Endpoint we will want the ability to force the lib to attempt to use the features we choose, and override the automatic detection

include/libebpf.h
tabell: I agree with @nicholasberlin here - I think it makes more sense for the library to manage the allocation of its own context, which requires it to return a pointer to pointer.

include/libebpf.h
mattnite: I agree, I think that we should provide the user with a single, ordered stream of events. My patch above would then just be:
```c
struct ebpf_context;

// placeholder
struct ebpf_event;

int ebpf__get_event_stream_fd(
        struct ebpf_context *ctx);

int ebpf__read_event(
        struct ebp_context *ctx,
        struct ebpf_event *out);
```

include/libebpf.h
fntlnz: Fine for me!

include/libebpf.h
```diff
@@ -59,7 +59,7 @@ struct ebpf_event {
  * of the mapping in the context */
 typedef struct ebpf_event_ctx;
 
-typedef int (*event_handler_fn)(struct ebpf_event *event, size_t);
+typedef int (*ebf_event_handler_fn)(struct ebpf_event *event, size_t);
```
leodido, line 62: Typo? It lacks a `p` :)

nicholasberlin:
```suggestion
typedef int (*ebpf_event_handler_fn)(struct ebpf_event *event, size_t);
```

include/libebpf.h
nicholasberlin: This PR is getting crowded with comments, perhaps we should close this and start fresh with our latest ideas.

m-sample: @mattnite naming tweak suggestion - simpler, no semantic implications

```
int ebpf__get_event_stream_fd(
        struct ebpf_context *ctx);
```
to 
```
int ebpf__get_event_fd(
        struct ebpf_context *ctx);
```

norrietaylor: Streaming implies ordered, lossless transmission and queuing. Think socket stream vs datagram. With the perf buffers, it probably isn't even possible to emulate a streaming interface. I would keep `stream` out of the name.

m-sample: I like the flexibility of something like this:

```C
int ebpf__get_event_fds(
        struct ebpf_context *ctx, u64 options, int **fds,  int* size);
```

It allows it to be used now with ringbuffer (one fd returned, time ordered events)

it allows those using this API that pay attention to the fd array size to work just fine with a perf buf backed implementation (e.g. perf-buf per core, no ordering guarantees)

It allows for options to force/mandate specific behavior - e.g. must use perf bufs even if ringbufers are avail (testing), or must be ordered in which case a sorting thread could be started if nec and one fd holding the sorted events could be returned via fds output param

it allows re-use and using your own allocator if you provide a non null *fds and non-zero *size and falling back to malloc if both are zero (error return if provided fds not big enough to avoid complexity of knowing when caller has to free(*fds))

include/libebpf.h
m-sample: how about something styled after perf_buffer__consume_buffer() in libbpf?  Perhaps the callback is registered with the ctx. and its return value, like the libbpf callback dictates whether it is called for every available event or returns before. The code around the calls to the callback takes care of advancing the perf_buf or ring buffer fd.

```C

// Calls ctx.cb to process available messages from the fd. 
// ctx knows if perf_buffer__consumer_buffer or ringbuffer__consume_buffer
// or the library-provided ordering fd over perf_bufs consume fcn must be called
int ebpf__consume_buffer(struct ebpf_event_ctx *ctx, int fd)
```

include/libebpf.h
ferullo: I love the flexibility of this.

include/libebpf.h
tabell: @m-sample I like the idea of pre-registering the callback instead of passing it to the read function but it would impose an additional requirement on the context to have a list of open fds with associated callbacks, right?

include/libebpf.h
m-sample: ts and type appear to represent common fields to each of our perf events or ringbuffer events that we create (ie are added by our BPF probe code). Presumably we'd have another header file with specific event structs to cast these into, based on the event type -- so Dan's original comment of unsigned char[0] as a marker for those additional fields seems to make sense?

include/libebpf.h
tabell: @m-sample I'm not clear how this fits with the model implied by the `ebpf_event__buffer_new` prototype that reads the options and opens the correct type of buffer. I was under the impression that by the time `ebpf__get_event_fds` is called, the decision of whether to use perf or ringbuf has already been made, and thus I don't understand why we need to pass `options` into `ebpf__get_event_fds`

include/libebpf.h
tabell: @m-sample if the caller allocates the context then it could no longer be opaque, right? I don't necessarily have a problem with that paradigm but it does mean we would be separating the allocation from the initialiization. If we go that direction we might want a different name than `buffer_new` - maybe `ebpf_event__buffer_init`?

fntlnz:
Rebased with main

include/libebpf.h
fntlnz: I'm pushing a change to fix the events to be a power of two since this is the way most people seem to like. We will see if 64 elements is enough while working on it.

include/libebpf.h
fntlnz: @m-sample the current `ebpf_event` is made of three members: ts, type and data.

ts and type are common, we use them for ordering and filtering, data is implementation dependent.

If we need anything else that should be at this level we probably want to add it to `ebpf_event` instead of having another opaque container in my opinion.

```
struct ebpf_event {
    uint64_t ts;
    uint64_t type;
    void *data;
};
```

include/libebpf.h
fntlnz: @m-sample @ferullo @tabell I thought we agreed on having a single file descriptor regardless the implementation. 
Also, since context is really an opaque data structure and implementation dependent, could the implementation just rely on that for this kind of things instead of having more fields?
Am I missing something?

include/libebpf.h
fntlnz: This is outdated, the api changed. Marking this as resolved

include/libebpf.h
fntlnz: @m-sample  Pre-registering the callback is the thing I was also advocating for in the initial version here [`d8cca9d` (#38)](https://github.com/elastic/ebpf/pull/38/commits/d8cca9d8d84da7f20dcb8aee42d17f57a1b5275e#diff-eaa9ba7b92779d1b656299e6902b1a73cf94074cb01e19ba526936a649a6290eR52)

However, passing the necessary scope to that callback without allocating a global variable is not really feasible.

In C++ one would use a lambda or an `std::function`.

We could use closures but they are a language extension and not a standard, so:

- clang has blocks https://clang.llvm.org/docs/BlockLanguageSpec.html
- gcc has closures https://gcc.gnu.org/onlinedocs/gcc/Nested-Functions.html

include/libebpf.h
fntlnz: Would you be able to provide an example on how this would work?

This is still confusing to me

include/libebpf.h
fntlnz: When writing this

```
struct ebpf_event {
    uint64_t ts;
    uint64_t type;
    void *data;
};
```
I didn't take in account the fact that while in the perf buffer we can allocate some kind of heap using a map with the ring buffer it's way more practical to just allocate the memory directly in the event itself.

So I made a change to accept @ferullo suggestion, Thanks Dan!

Commit here [`cb8fa5d` (#38)](https://github.com/elastic/ebpf/pull/38/commits/cb8fa5d0696b3cd1001883944a8e34c859763eca)


We can discuss more this approach, but it is a starting point, I got generic events flowing using `ebpf_event` as header working with this.

fntlnz:
@tabell I added some ifdefs to detect whether we are in kernel or user space to the header.

This is useful because for example the event type needs to be written by kernel space, and consumed by user space.

commit: [`085cafb` (#38)](https://github.com/elastic/ebpf/pull/38/commits/085cafb3f4c8aa22d4f0fc74fcf5877cb828e599)

include/libebpf.h
```diff
@@ -0,0 +1,103 @@
...
+#define LIBEBPF_H_
+
+#if !defined(__KERNEL__) && !defined(__VMLINUX__)
+#include <stdint.h>
+#include <sys/types.h>
+#endif // !defined(__KERNEL__) && !defined(__VMLINUX__)
+
+/* Kernel Features */
+enum ebpf_kernel_features {
+    EBPF_KERNEL_CAPABLE_BPF = (1 << 0),
```
rhysre, line 22: Can we completely avoid the use of the terms `capable` and `capability`? IMO there's a lot of potential here for confusion with Linux capabilities. `supports` or `feature` would work just as well.

include/libebpf.h
```diff
@@ -0,0 +1,103 @@
...
+    EBPF_KERNEL_CAPABLE_BTF = (1 << 2),
+};
+
+/* Events fetching and manipulation */
+enum ebpf_event_type {
+    EBPF_EVENT_PROCESS_FORK = (1 << 1),
+    EBPF_EVENT_PROCESS_EXEC = (1 << 2),
+    EBPF_EVENT_PROCESS_EXIT = (1 << 3),
+    EBPF_EVENT_PROCESS_TTY_OUTPUT = (1 << 4),
+    EBPF_EVENT_FILE_CREATE = (1 << 5),
```
rhysre, line 33: In the very long term, I'm slightly concerned about us going over 64 events. I can see the `ebpf_kernel_features` enum not growing very much, but intuitively I feel like the number of events would grow faster with time. 

Right now I don't see us getting close to 64, but this is the type of thing I could see causing problems in 5 years if we have other people using this API externally and need to keep it stable.

include/libebpf.h
```diff
@@ -0,0 +1,103 @@
...
+ * struct ring_buffer,
+ *
+ * we need a way to map from capability/event sets to probe sets,
+ * and we could store the cap/event sets directly or cache the result
+ * of the mapping in the context */
+struct ebpf_event_ctx;
+
+typedef int (*ebpf_event_handler_fn)(struct ebpf_event *event, size_t);
+
+typedef uint64_t ebpf_cap_set;
```
rhysre, line 62: ditto on changing the `cap`/`capability` terminology to avoid confusion here

include/libebpf.h
rhysre: +1 On an example. This raised a flag for me too. What features would we want to manually force if feature detection turned them off?

include/libebpf.h
```diff
@@ -0,0 +1,103 @@
...
+ * does not need to know which probes to load */
+typedef uint64_t ebpf_event_set;
+
+/* allocates a new context based on requested events and capabilities.
+ *
+ * returns a positive int that represents an fd, which
+ * can be used with epoll on success. returns -1 and
+ * sets errno on failure
+ */
+int ebpf_event__buffer_new(
```
rhysre, line 78: I believe this should be `ebpf_event_buffer__new`.

include/libebpf.h
```diff
@@ -0,0 +1,103 @@
...
+ * func when fd is readable. I think passing a handler
+ * function is more versatile than passing a buffer?
+ * no strong feelings on that. The handler will likely
+ * dispatch one of another layer of callbacks switched
+ * on event type.
+ *
+ * We should also discuss blocking/nonblocking semantics
+ * here. I think this function could return -1 and set
+ * errno to -EAGAIN if the context is in a nonblocking
+ * state but no data is available. */
```
rhysre, line 94: I don't see a circumstance (given what we want to do currently) in which we'd need a nonblocking mode. If we're always calling this after `epoll` notifies us of data being ready, just calling it in blocking mode should suffice. 

This seems like something we can add later if we need it.

rhysre (commented):
Commented on a few things that stood out to me.

This has been said, but this PR is becoming a bit of an impossible-to-read mosh pit of comments. 

Given that a fair bit of feedback has been addressed, we may just want to start the implementation side of things and iterate a bit more on the API as it becomes necessary. I don't forsee massive changes to the API at this point given how much feedback has already been voiced.

tabell (approved):

include/libebpf.h
fntlnz: Yeah that's correct! Feel free to update this before merging, or we can do that in another PR.

Good catch

nicholasberlin (approved):

PR #39, opened 2021-11-15
docs(proposals): brain dump on feature detection for eBPF

leodido:
Hello all! 👋 

Opening this proposal about feature detection.

It is a first, initial draft that dumps everything I have on my brain about feature detection for eBPF.

By analyzing the problem, I spec'ed out some initial (and simple) candidate API designs.

I'd like to get to know your feedback.

Think of this as a discussion starter.

---

The document contains images you don't wanna miss so it's worth clicking on "view file" to get to see them!

### DIFF
--- a/docs/proposals/feature-detection.md
+++ b/docs/proposals/feature-detection.md
@@ -95,7 +95,7 @@ Then we could simply include such C header and use it in our programs.
 
 ***Pros***:
 
-- It is a the approach more common used in eBPF land (as of today at least)
+- It is the approach more commonly used in eBPF land (as of today at least)
 
 ### Runtime
 
@@ -114,10 +114,11 @@ bool has_helper_for_prog(string helper, u32 prog_type, bool unprivileged, u32 if
 Such an API would be called like so (pseudo-code):
 
 ```c
-has_map(BPF_MAP_TYPE_LPM_TRIE, false);
-has_prog(BPF_PROG_TYPE_LSM, false);
-has_prog(BPF_PROG_SOCKET_FILTER, true);
-has_helper_for_prog("bpf_get_stackid", BPF_PROG_TYPE_TRACING, false);
+has_map(BPF_MAP_TYPE_LPM_TRIE, false, 0);
+has_prog(BPF_PROG_TYPE_LSM, false, 0);
+has_prog(BPF_PROG_TYPE_SCHED_CLS, false, 2/*eth0*/);
+has_prog(BPF_PROG_SOCKET_FILTER, true, 0);
+has_helper_for_prog("bpf_get_stackid", BPF_PROG_TYPE_TRACING, false, 0);
 ```
 
 Some observations.
@@ -126,7 +127,7 @@ Some observations.
 
 This API has 2 targets: **kernel** or **device**.
 
-The `if_idx` argument is only relevant when greather than 0, pointing to a device interface (device target).
+The `if_idx` argument is only relevant when greater than 0, pointing to a device interface (device target, eg. 1 being `lo` iface, 2, etc.).
 Otherwise, and by default, the API probes the kernel (kernel target).
 
 **Usage observations**:
--- a/docs/proposals/feature-detection.md
+++ b/docs/proposals/feature-detection.md
@@ -1,0 +1,255 @@
+# Feature detection for eBPF
+
+## Privileges
+
+To build a machinery that does full feature detection requires specific **privileges**.
+
+Namely, it requires `CAP_BPF`, `CAP_NET_ADMIN`, and `CAP_PERFMON`.
+If the kernel does not know about `CAP_BPF`, such a machinery requires `CAP_SYS_ADMIN`.
+
+Anyways, detecting unprivileged eBPF features can be done without the full privileges (it would require `libcap`).
+
+So, the first aspect to keep in mind is: **privileges**.
+
+## Features
+
+Then, there is the **feature** aspect.
+
+We should also distinguish between Kernel configurations and eBPF features. Often, the latter depends on the former.
+
+I envision the following **categories** of checks:
+
+1. system configurations (requires procfs)
+   1. check unprivileged eBPF is disabled (`/proc/sys/kernel/unprivileged_bpf_disabled`)
+   2. check JIT
+      1. enable (`/proc/sys/net/core/bpf_jit_enable`)
+      2. harden (`/proc/sys/net/core/bpf_jit_harden`)
+      3. kallsysm (`/proc/sys/net/core/bpf_jit_kallsyms`)
+      4. limit (`/proc/sys/net/core/bpf_jit_limit`)
+   3. check kernel image (`/boot/config-*`, `/proc/config.gz`) for:
+      1. `CONFIG_BPF`, `CONFIG_BPF_SYSCALL`, `CONFIG_HAVE_EBPF_JIT`, `CONFIG_BPF_JIT`, `CONFIG_BPF_JIT_ALWAYS_ON`
+      2. `CONFIG_DEBUG_INFO_BTF`, `CONFIG_DEBUG_INFO_BTF_MODULES`
+      3. etc.
+2. syscall configuration
+   1. try to load a `BPF_PROG_TYPE_UNSPEC` with `bpf` syscall and grab the result
+3. supported program types
+   1. Use `bpf_probe_prog_type` of `libbpf` (present since v0.0.2) or equivalent implementation
+      1. It loads an eBPF program type (depending on the program type to probe) composed of 2 instructions (`MOV64_IMM(r0, 0)`, `EXIT_INSN()`)
+      2. Checks the outcome
+   2. For unprivileged case, check that `errno` is not `EPERM` (insufficient permissions)
+4. supported map types
+   1. Use `bpf_probe_map_type` of `libbpf` (present since v0.0.2) or equivalent implementation
+5. supported helper types by program type
+   1. Use `bpf_probe_helper` of `libbpf` (present since v0.0.2)
+      1. It loads an eBPF program (of the type for which to probe the helper/s) composed of 2 instructions (`EMIT_CALL(helper_id)`, `EXIT_INSN()`)
+      2. Checks the outcome
+   2. Only account for offloadable program types when targeting a device
+   3. Future:
+      1. notice that some helper functions emit dmesg messages (eg., `BPF_FUNC_trace_printk`), we may want to separate them from the rest
+      2. whether an helper is GPL-only or not: requires reading the `_proto` kernel structs
+6. others miscellaneous features
+   1. Detect whether programs up to `BPF_MAXINSNS` instructions are supported
+      1. Use `bpf_probe_large_insn_limit` of `libbpf` (present since v0.0.7) or equivalent implementation
+         1. Especially useful when targeting a device
+
+## Targets
+
+Finally, it is imporant noticing that the third and last aspect to consider is the **target** we are probing.
+
+It's important to distinguish whether these checks have to run against a device.
+
+We have 2 possible **targets** for this API: kernel or device.
+
+## API
+
+To me, we do have basically two possible API designs available (**not mutually exclusive since the second can generate the first**).
+
+### Preprocessing/compilation time
+
+We may build a machinery that runs steps 1-6 and it generates a **C header** containing macros/defines for the various system configurations and eBPF features
+detected on the current kernel/device target.
+
+Then we could simply include such C header and use it in our programs.
+
+```c
+#define HAS_BPF_SYSCALL
+// ...
+#define HAS_KPROBE_PROG_TYPE
+#define HAS_SCHED_CLS_PROG_TYPE
+// ...
+#define HAS_HASH_MAP_TYPE
+#define HAS_ARRAY_MAP_TYPE
+// ...
+#define PROG_TYPE_HAS_HELPER(prog_type, helper) \
+    PROG_TYPE_ ## prog_type ## __HAS_HELPER ## helper
+// ...
+#define PROG_TYPE_lsm__HAS_HELPER_bpf_tail_call 0
+```
+
+***Cons***:
+
+- Such a header would need to be generated on the target kernel
+  - I believe this is a **blocker** for us
+  - It makes no sense to talk about BTF and CO-RE if then we still need to generate headers for different target kernels
+- Very verbose, especially in case we end up defining macro variations for the unprivileged case too and for different targets.
+
+***Pros***:
+
+- It is a the approach more common used in eBPF land (as of today at least)
+
+### Runtime
+
+#### Stateless lower-level API
+
+We may build a API like the following (in pseudo-code):
+
+```c
+bool can_bpf(bool unprivileged, u32 if_idx);
+bool has_map(u32 map_type, bool unprivileged, u32 if_idx);
+bool has_prog(u32 prog_type, bool unprivileged, u32 if_idx);
+bool has_helper_for_prog(string helper, u32 prog_type, bool unprivileged, u32 if_idx);
+// ...
+```
+
+Such an API would be called like so (pseudo-code):
+
+```c
+has_map(BPF_MAP_TYPE_LPM_TRIE, false);
+has_prog(BPF_PROG_TYPE_LSM, false);
+has_prog(BPF_PROG_SOCKET_FILTER, true);
+has_helper_for_prog("bpf_get_stackid", BPF_PROG_TYPE_TRACING, false);
+```
+
+Some observations.
+
+**Targets**:
+
+This API has 2 targets: **kernel** or **device**.
+
+The `if_idx` argument is only relevant when greather than 0, pointing to a device interface (device target).
+Otherwise, and by default, the API probes the kernel (kernel target).
+
+**Usage observations**:
+
+Notice such an API is meant to be called from the loader of an eBPF program, not from within an eBPF program.
+
+Turns out,
+this means we may ending up writing different eBPF programs for the same feature with different implementation (eg., perf buffer vs ring buffer)
+and load one or the other depeding on the feature existence we detect before loading.
+
+I personally consider this a +1 because I do believe it would force us to write more self-contained eBPF programs without `ifdefs` or `if` chain hells.
+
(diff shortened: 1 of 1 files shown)


nicholasberlin:
I like the idea of having separate eBPF programs using different features. When there are a lot of `#ifdef`s in a single code base, the code reuse benefit starts to be overcome by the confusion added by mentally parsing `#ifdef`s.

If we decide on the runtime route, the code utilizing bpfhaz lib would have to know the requirements of what it's loading. The loading code, `bpfhaz` lib, and the eBPF programs would all need to be kept in sync. Is it possible to examine the eBPF .o's to know what is required?

Is there any harm in attempting to load the latest and greatest program set, and backing off to less feature rich sets until something successfully loads?

Is it possible to successfully load a program that won't work because a feature is missing?

fntlnz:
Thanks @leodido ! I like this proposal and I'd love to see it implemented.

I particularly like the fact that this is very flexible and that the API can be used to build other checks on top of it.

cc/ @ferullo @stanek-michal @tabell

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,255 @@
...
+// ...
+```
+
+Such an API would be called like so (pseudo-code):
+
+```c
+has_map(BPF_MAP_TYPE_LPM_TRIE, false);
+has_prog(BPF_PROG_TYPE_LSM, false);
+has_prog(BPF_PROG_SOCKET_FILTER, true);
+has_helper_for_prog("bpf_get_stackid", BPF_PROG_TYPE_TRACING, false);
```
leodido, line 120:
```suggestion
has_map(BPF_MAP_TYPE_LPM_TRIE, false, 0);
has_prog(BPF_PROG_TYPE_LSM, false, 0);
has_prog(BPF_PROG_TYPE_SCHED_CLS, false, 2/*eth0*/);
has_prog(BPF_PROG_SOCKET_FILTER, true, 0);
has_helper_for_prog("bpf_get_stackid", BPF_PROG_TYPE_TRACING, false, 0);
```

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,255 @@
...
+***Cons***:
+
+- Such a header would need to be generated on the target kernel
+  - I believe this is a **blocker** for us
+  - It makes no sense to talk about BTF and CO-RE if then we still need to generate headers for different target kernels
+- Very verbose, especially in case we end up defining macro variations for the unprivileged case too and for different targets.
+
+***Pros***:
+
+- It is a the approach more common used in eBPF land (as of today at least)
```
leodido, line 98:
```suggestion
- It is the approach more commonly used in eBPF land (as of today at least)
```

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,255 @@
...
+has_helper_for_prog("bpf_get_stackid", BPF_PROG_TYPE_TRACING, false);
+```
+
+Some observations.
+
+**Targets**:
+
+This API has 2 targets: **kernel** or **device**.
+
+The `if_idx` argument is only relevant when greather than 0, pointing to a device interface (device target).
```
leodido, line 129:
```suggestion
The `if_idx` argument is only relevant when greater than 0, pointing to a device interface (device target, eg. 1 being `lo` iface, 2, etc.).
```

leodido (commented):
Fix some typos

leodido:
> I like the idea of having separate eBPF programs using different features. When there are a lot of `#ifdef`s in a single code base, the code reuse benefit starts to be overcome by the confusion added by mentally parsing `#ifdef`s.

Totally agree.

> If we decide on the runtime route, the code utilizing bpfhaz lib would have to know the requirements of what it's loading. The loading code, `bpfhaz` lib, and the eBPF programs would all need to be kept in sync. Is it possible to examine the eBPF .o's to know what is required?

It is very possible to understand what we need by parsing the ELF. It's what `libbpf` already does.
Furthermore, if we end up generating skeletons and loading them we could simply read such objects (so no parsing, just navigating objects) and understand the map types and program types.
Detecting the helpers can be done by iterating over the instructions and looking up for `call ID` instructions, from the ID you obtain the helper name looking it up in BTF info.

So it's doable and also ~ already done depending on the approach we end up using for loading the programs. :)

> Is there any harm in attempting to load the latest and greatest program set, and backing off to less feature rich sets until something successfully loads?

Hmmm, the only type of harm (that I can think of rn) this could cause is described by the example below.
 
> Is it possible to successfully load a program that won't work because a feature is missing?

Yes.

The first example that comes to my mind is `bpf_printk`.

Newer versions (on newer kernels) of this helper accept more than one "%s" qualifier in the format, it understands "%pI4" qualifier, etc.
If you use those new features on an older kernel, the eBPF verifier accepts them, and the program loads but does not work when reaching that instruction...

Notice, I omitted in the current proposal, how to detect "which version" of the helper we have. That's for a follow-up iteration (wanted to keep it simple) :)

My memory tells me that I met other similar cases but I can wrap my head to remember them at the moment.

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+We should also distinguish between Kernel configurations and eBPF features. Often, the latter depends on the former.
+
+I envision the following **categories** of checks:
+
+1. system configurations (requires procfs)
+   1. check unprivileged eBPF is disabled (`/proc/sys/kernel/unprivileged_bpf_disabled`)
+   2. check JIT
+      1. enable (`/proc/sys/net/core/bpf_jit_enable`)
+      2. harden (`/proc/sys/net/core/bpf_jit_harden`)
+      3. kallsysm (`/proc/sys/net/core/bpf_jit_kallsyms`)
```
leodido, line 27: Typo

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+
+Ipothetically, a top layer API, rather than issuing a call instruction to test the availability of a given helper,
+may simply use BTF knowing it's available.
+
+Fallbacking to the one based on instructions when BTF is not there.
+
+Notice that the same mechanism can be used for map and program types too.
+Notice also that this same mechanism can be used also for non-BPF things.
+
+#### Stateful higher-level API
```
leodido, line 238: Will expand further on this ASAP

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+      1. notice that some helper functions emit dmesg messages (eg., `BPF_FUNC_trace_printk`), we may want to separate them from the rest
+      2. whether an helper is GPL-only or not: requires reading the `_proto` kernel structs
+6. others miscellaneous features
+   1. Detect whether programs up to `BPF_MAXINSNS` instructions are supported
+      1. Use `bpf_probe_large_insn_limit` of `libbpf` (present since v0.0.7) or equivalent implementation
+         1. Especially useful when targeting a device
+
+## Targets
+
+Finally, it is imporant noticing that the third and last aspect to consider is the **target** we are probing.
```
rockdaboot, line 57: Typo: imporant -> important

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+
+***Pros***:
+
+- It is the approach more commonly used in eBPF land (as of today at least)
+
+### Runtime
+
+#### Stateless lower-level API
+
+We may build a API like the following (in pseudo-code):
```
rockdaboot, line 104: a -> an

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+
+The `if_idx` argument is only relevant when greater than 0, pointing to a device interface (device target, eg. 1 being `lo` iface, 2, etc.).
+Otherwise, and by default, the API probes the kernel (kernel target).
+
+**Usage observations**:
+
+Notice such an API is meant to be called from the loader of an eBPF program, not from within an eBPF program.
+
+Turns out,
+this means we may ending up writing different eBPF programs for the same feature with different implementation (eg., perf buffer vs ring buffer)
```
rockdaboot, line 138: -> implementations

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+and load one or the other depeding on the feature existence we detect before loading.
```
rockdaboot, line 139: -> depending

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+2. populate the `rodata` section (one or more `const` variables in the eBPF program)
+3. write the eBPF program to react to those `const` variables at runtime
+
+This way we may write a single eBPF program that, depending on the conditions on those variables, uses one feature or the other.
+Notice that global data sections for eBPF programs requires kernel 5.2. But the same overall behavior can be achieved with plain eBPF maps.
+
+**Unprivileged probing**:
+
+If the API user wants to explicitly probe feature existence for unprivileged use case while it has the full privileges,
+the API should in that case lower the privileges is using. Then, proceed as normal.
```
rockdaboot, line 156: is -> it is  or just drop `is using`.

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+
+Also the reverse can happen. An API user with low privileges that wants to probe eBPF features requiring more privileges.
+
+In such case, I advice to just return `false` for the specific feature being probed even if we can't really know if it would be available with more privileges or not.
```
rockdaboot, line 160: -> such a case

docs/proposals/feature-detection.md
syscll:
```suggestion
this means we may end up writing different eBPF programs for the same feature with different implementation (eg., perf buffer vs ring buffer)
```

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+
+Anyways, in case we would like to use this API in an eBPF program, we could nevertheless.
+By following these steps:
+
+1. detect the feature we need for the eBPF program we are about to load
+2. populate the `rodata` section (one or more `const` variables in the eBPF program)
+3. write the eBPF program to react to those `const` variables at runtime
+
+This way we may write a single eBPF program that, depending on the conditions on those variables, uses one feature or the other.
+Notice that global data sections for eBPF programs requires kernel 5.2. But the same overall behavior can be achieved with plain eBPF maps.
```
syscll, line 151:
```suggestion
Notice that global data sections for eBPF programs require kernel 5.2. But the same overall behavior can be achieved with plain eBPF maps.
```

docs/proposals/feature-detection.md
syscll, line 156:
```suggestion
the API should in that case lower the privileges it's using. Then, proceed as normal.
```

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+
+**Unprivileged probing**:
+
+If the API user wants to explicitly probe feature existence for unprivileged use case while it has the full privileges,
```
syscll, line 155:
```suggestion
If the API user wants to explicitly probe feature existence for unprivileged use cases while it has the full privileges,
```

docs/proposals/feature-detection.md
syscll, line 160:
```suggestion
In such cases, I advise to just return `false` for the specific feature being probed even if we can't really know if it would be available with more privileges or not.
```

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+
+**System configs**:
+
+Point 1 of the above list.
+
+This step needs procfs to parse the sysctl knobs that govern some eBPF features.
+
+Also, it needs to parse the kernel config.
+
+To avoid writing a parser for the Kernel config, a easier/cleaner approach
```
syscll, line 170:
```suggestion
To avoid writing a parser for the Kernel config, an easier/cleaner approach
```

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+would be to write an eBPF program that uses **Kconfig externs** to probe the Kernel configs and then stores such values in a pinned map.
+By loading this probe at the start of our future programs pipeline, we can avoid parsing multiple times the Kernel config (read below).
+
+But such approach requires anyways the Kernel config (eg., `/proc/config.gz`).
```
syscll, line 174:
```suggestion
But such approaches require the Kernel config (eg., `/proc/config.gz`).
```

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+To do so, we'd need to check various things at system configuration level (point 1), for example whether BTF is enabled or not first of all, and then at other levels too (`BPF_PROG_TYPE_LSM`).
+
+So, it could make sense to build a **stateful layer on top of this stateless API**.
+
+An initialization step that grabs (and stores/caches) the kernel config values to avoid having to parse for them again (disregarding how we implement this, see above).
+
+There are cases in which it can be pretty useful to have all the needed values in place.
+In fact, it is also very common that a Kernel config depends on other Kernel configs (see `config BPF_LSM` for example, [link](https://github.com/torvalds/linux/blob/5d6ab0bb408ffdaac585982faa9ec8c7d5cc349f/kernel/bpf/Kconfig#L77)).
+
+So, we can define DAGs (or FSMs) to declaratively express feature detection at **an higher granularity level**.
```
syscll, line 203:
```suggestion
So, we can define DAGs (or FSMs) to declaratively express feature detection at **a higher granularity level**.
```

docs/proposals/feature-detection.md
```diff
@@ -0,0 +1,256 @@
...
+
+```c
+if (bpf_core_enum_value_exists(enum bpf_func_id, BPF_FUNC_ringbuf_reserve)) {
+    // use bpf_ringbuf_reserve()
+} else {
+    // fallback to bpf_perf_event_output()
+}
+```
+
+Ipothetically, a top layer API, rather than issuing a call instruction to test the availability of a given helper,
```
syscll, line 230:
```suggestion
Hypothetically, a top layer API, rather than issuing a call instruction to test the availability of a given helper,
```

syscll (changes requested):
Really great document! Sorry for the nits.

rhysre:
I'll echo @nicholasberlin 's comment about not going the macro-generation route to avoid lots and lots of preprocessor defines. That was the way we did things at Cmd and, although it works well enough for our current use cases, I can see it really starting to clutter things up in the future.

The only minor concern I would have if we went the other way (and this could very well be unfounded) would be about us having to maintain an unreasonable number of probe codebases. Consider, for instance, four BPF features, A, B, C and D, with kernels existing that have features A, AB, ABC and ABCD. If those features were all relevant to us, we might have to maintain four separate sets of slightly different probes, keeping them in sync when we make new changes and add new features. I'd wager it would be pretty easy to cut the number of probe codebases down (eg. maybe we just do two sets, A and ABCD if we figure the number of kernels with AB and ABC are rather low), but some food for though nonetheless.

Regardless, I like that approach over macro generation.

> Is there any harm in attempting to load the latest and greatest program set, and backing off to less feature rich sets until something successfully loads?
>
> Is it possible to successfully load a program that won't work because a feature is missing?

I think this is definitely worth exploring, but would be hesitant about it. While I can't think of any examples offhand (besides the `printk` issue Leo brought up) I don't think there's a formal guarantee that any program requiring feature X will fail to load because feature X is missing. Off the top of my head I'd prefer explicit checks for features in the endpoint codebase before we try to load something.

leodido:
> I'll echo @nicholasberlin 's comment about not going the macro-generation route to avoid lots and lots of preprocessor defines. That was the way we did things at Cmd and, although it works well enough for our current use cases, I can see it really starting to clutter things up in the future.

So, it seems we all agree on the fact that the macro-generation approach is a no-go for us! 
I listed it because I wanted to give a complete view of the possibilities so that we all could make an informed decision. :)

> The only minor concern I would have if we went the other way (and this could very well be unfounded) would be about us having to maintain an unreasonable number of probe codebases. Consider, for instance, four BPF features, A, B, C and D, with kernels existing that have features A, AB, ABC and ABCD. If those features were all relevant to us, we might have to maintain four separate sets of slightly different probes, keeping them in sync when we make new changes and add new features. I'd wager it would be pretty easy to cut the number of probe codebases down (eg. maybe we just do two sets, A and ABCD if we figure the number of kernels with AB and ABC are rather low), but some food for thought nonetheless.

Yes. Theoretically, we could tend towards an explosion of programs.
This is one of the reasons I envision two APIs (as written down in the document): one to check the existence of single features (eg., is `BPF_PROG_TYPE_LSM` available?), and another (built on top of the first) checking whether the whole feature is actually usable (see the FSM example for checking BPF LSM support).

In fact, having `BPF_PROG_TYPE_LSM` (the "single" feature) does not guarantee we can run BPF LSM programs.

So, I'm fairly confident we will end up checking for "whole" features and the number of resulting probes will depend on them, not on single features. This reduces a lot the number of different probes to write/support and to keep in sync.

Furthermore, it's worth mentioning that in some (edge) cases we could also adopt a mixed approach.
Ie., moving the feature detection from the loading phase to the eBPF ELF with macros for single features, so as to reduce again the number of possible resulting probes.

My only concern for this mixed approach is it can easily degenerate into a mess (like if we went for the macro approach, or worse) in case we do not set up a strict and clear set of rules for when to use it and when not.

Overall, I believe that the combinatorial explosion is unlikely because of the reasons stated above.

> Regardless, I like that approach over macro generation.

YaY!

docs/proposals/feature-detection.md
leodido:
```suggestion
Finally, it is important to notice that the third and last aspect to consider is the **target** we are probing.
```

docs/proposals/feature-detection.md
leodido:
```suggestion
We may build an API like the following (in pseudo-code):
```

docs/proposals/feature-detection.md
leodido:
```suggestion
this means we may ending up writing different eBPF programs for the same feature with different implementations (eg., perf buffer vs ring buffer)
```

fntlnz (approved):
I like this proposal, we can use this as a starting point for an implementation.

LGTM

rhysre (approved):

Issue #43, opened 2021-12-14
Discussion: Variable length or fixed-length wire format for events


rhysre:
When sending events up via perfbuffers/ringbuffers we need to decide on whether we're going to use a variable-length or a fixed-length binary format.

Take a process execution for example. Using a fixed-length format, it might look something like this:

```c
struct event_process_exec {
    uint32_t tid;         // Thread ID
    uint32_t tgid;        // Thread group ID
    uint32_t pgid;        // Process group ID
    uint32_t sid;         // Session ID
    uint32_t ruid;        // Real user ID
    uint32_t euid;        // Effective user ID
    char path[PATH_MAX];  // Path to binary being executed
    char pwd[PATH_MAX];   // Current working directory
    // Perhaps more fields down here
} __attribute__((packed));
```

Having a single, fixed-length struct like this is incredibly simple and makes for easy parsing, but is space inefficient when considering the variable-length components of the above struct (i.e. the paths). `PATH_MAX` on Linux is commonly 4096, if `path` and `pwd` were both 10 bytes, we'd be shoving ~8KB in the perfbuffer/ringbuffer that doesn't need to be there.

We could get around this with a variable length format instead that attaches a header to each parameter specifying its length. This would be much more space efficient at the cost of added complexity in the parsing code (and extra CPU work for parsing, but we're really getting into premature optimization land with that one).

My initial thought was to go with the variable length format due to the space efficiency. I worried about a machine generating an enormous amount of events, overflowing the perf/ring buffers and us losing events. The variable-length idea however, doesn't seem to be a very good fit for the BPF ringbuffer's reserve/commit API. Given that you have to `bpf_ringbuf_reserve` a chunk of memory _in advance_ (i.e. before we know how long the event would actually be in the variable length format), we'd have to reserve (and later submit) a spot in the ringbuffer that's the maximum possible event size anyways, thus defeating the purpose of a variable-length format.

We could use `bpf_ringbuf_output` instead, which allows us to specify a size, but this would require us ditching the reserve/commit API, which is a nice advantage of the ringbuffer. This however, doesn't apply to perfbuffers of course. If we add a perfbuffer implementation in the future, perhaps we might want to use a fixed-length format for ringbufs and a variable-length format for perfbufs? Just riffing here to facilitate discussion.

[...]

PR #47, opened 2021-12-16
Add basic implementation of "libebpf" header

rhysre:
This PR adds a basic implementation to go along with the file found in include/libebpf.h.

The library was renamed to "LibEbpfEvents" (now found under `non-GPL/LibEbpfEvents`) to be more explicit about what it does. It currently uses a single set of ringbuffer probes regardless of the features/events requested and only supports the BPF ringbuffer. A perfbuffer implementation can be added at a later point if necessary.

The actual wire protocol has been factored out of LibEbpfEvents and into a dual-licensed header at `GPL/Events/EbpfEventProto.h`. This makes for easy including in the probes (no janky `#ifdef`s to exclude userspace code when including it in the probes) and easy dual-licensing of only the part of the events system that needs to be dual-licensed (the wire format).

The actual wire-format is the "fixed-length" protocol discussed in #43. This can be changed to variable-length in the future if necessary, but is the simplest thing to do for now to get events flowing.

This commit also adds very basic process fork and exec events to the existing events probe and renames the FileEvents probe to simply Probe. We can split different types of probes off later, but for this iteration, the simplest thing to do seems to be just to have all event types in a single probe.

The EventsTrace program at `non-GPL/EventsTrace` logs the new process events and should serve as a good example of how to use the new implementation.

CC: @fntlnz @leodido @elastic/awp-sensor 

Looking forward to the discussion 🚀


non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -0,0 +1,92 @@
...
+out_destroy_probe:
+    Probe_bpf__destroy((*ctx)->probe);
+    free(*ctx);
+    return err;
+}
+
+int ebpf_event_ctx__next(
+        struct ebpf_event_ctx *ctx)
+{
+    int consumed = ring_buffer__poll(ctx->ringbuf, 0);
```
fntlnz, line 85: Polling the ring buffer with no timeout leads to a busy CPU on my environment.


```
%CPU %MEM CMD
97.9  0.0 ./non-GPL/EventsTrace/EventsTrace
```

I'm not sure that keeping a full CPU busy is what we want.

```suggestion
    int consumed = ring_buffer__poll(ctx->ringbuf, 10);
```

Adding a default timeout `10ms` makes things better already:


```
%CPU %MEM CMD
 0.7  0.0 ./non-GPL/EventsTrace/EventsTrace
```


This is with `100ms`

```suggestion
    int consumed = ring_buffer__poll(ctx->ringbuf, 100);
```


```
%CPU %MEM CMD
 0.2  0.0 ./non-GPL/EventsTrace/EventsTrac
```


We can probably make this configurable as well.

Wdyt @rhysre ?

fntlnz (commented):
Good job @rhysre 

Just a couple of things and then I think we will be g2g

Thanks for working on this

.gitignore
```diff
@@ -5,3 +5,7 @@
 /non-GPL/HostIsolation/KprobeConnectHook/build/
 /non-GPL/HostIsolationMapsUtil/build/
 /non-GPL/TcLoader/build/
+
+# Vim swapfiles
+[._]*.s[a-w][a-z]
```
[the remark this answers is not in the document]
rhysre: Good call, removed them.

cmake/modules/BPF.cmake
```diff
@@ -9,7 +9,7 @@
 # Tools and defines
 set(CLANG "clang")
 set(LLC "llc")
-set(BPFTOOL "/usr/bin/bpftool") # TODO: the one in toolchain is old, need to update
+set(BPFTOOL "/usr/sbin/bpftool") # TODO: the one in toolchain is old, need to update
```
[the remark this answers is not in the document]
rhysre: Done

GPL/Events/Helpers.h
```diff
@@ -52,11 +48,58 @@ static __always_inline void *ebpf_event__new(void *ringbuf, uint64_t event_type)
...
-     
+
+    event->type = event_type;
+    event->ts = bpf_ktime_get_ns();
+
 out:
     return event;
 }
 
+static __always_inline int ebpf_event_file_path__from_dentry(struct ebpf_event_file_path *dst, struct dentry* src)
```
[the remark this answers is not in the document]
rhysre: I actually meant to delete `FileEvents.h` and just put all helpers into `Helpers.h`. Forgot to delete FileEvents.h though.

IMO at this point we can just put all helpers into a single `Helpers.h` file. When we start getting multiple probes and things get more complicated, we can split the helpers into multiple files. Thoughts?

non-GPL/EventsTrace/CMakeLists.txt
```diff
@@ -5,10 +5,13 @@
...
 add_executable(EventsTrace EventsTrace.c)
-add_dependencies(EventsTrace FileEvents_skeleton libbpf libelf)
-target_link_libraries(EventsTrace ${LIBBPF_LIB} ${LIBELF_LIB} -lz)
+add_dependencies(EventsTrace LibEbpfEvents Probe_skeleton libbpf libelf)
+target_link_libraries(EventsTrace LibEbpfEvents ${LIBBPF_LIB} ${LIBELF_LIB} -lz)
+
 target_include_directories(EventsTrace PRIVATE
-    "${CMAKE_SOURCE_DIR}/include/"
-    "${CMAKE_SOURCE_DIR}/GPL/Events"
+    "${PROJECT_SOURCE_DIR}/include/"
```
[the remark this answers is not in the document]
rhysre: done

GPL/Events/Probe.bpf.c
```diff
@@ -35,22 +34,19 @@ int BPF_PROG(security_path_unlink_exit, const struct path *dir, struct dentry *d
...
     if (!event)
     {
         // todo(fntlnz): fentry cannot return anything but zero, handle error here
         goto out;
     }
 
-    
     edata = (struct ebpf_event_file_delete_data *)event->data;
-    ebpf_event_file_delete_data__set_pid(edata, bpf_get_current_pid_tgid() >> 32);
+    edata->pid = bpf_get_current_pid_tgid() >> 32;
```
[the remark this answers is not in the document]
rhysre: That was in FileEvents.h which I meant to delete.

GPL/Events/CMakeLists.txt
```diff
@@ -20,45 +20,48 @@
...
     "${TARGET_INCLUDE_DIR}"
     "${GTEST_INCLUDE}")
 
-add_dependencies(BPFFileEventsTests BPFFileEvents)
+add_dependencies(BPFEventsTests BPFEventProbe)
 
-bpf_skeleton(FileEvents)
-add_dependencies(FileEvents_skeleton BPFFileEvents)
+bpf_skeleton(Probe)
+add_dependencies(Probe_skeleton BPFEventProbe)
```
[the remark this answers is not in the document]
rhysre: Done.

GPL/Events/CMakeLists.txt
```diff
@@ -20,45 +20,48 @@
...
                 -D__KERNEL__
                 -D__TARGET_ARCH_${ARCH_TRUNC}
                 -D__${ARCH}__
+                -I${TARGET_INCLUDE_DIR}
                 -I${LIBBPF_UAPI_INCLUDE_DIR}
                 -I${VMLINUX_INCLUDE_DIR}
-                -I${FILEEVENTS_SOURCE_DIR}
-                -I${CMAKE_SOURCE_DIR}/include)
+                -I${EVENTS_PROBE_SOURCE_DIR}
+                -I${PROJECT_SOURCE_DIR}/include)
```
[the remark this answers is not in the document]
rhysre: Done.

non-GPL/LibEbpfEvents/LibEbpfEvents.c
rhysre: Good call. I added an arg to `event_ctx__next`. I was looking at this from the perspective of non-blocking application that uses the epoll fd, but of course EventsTrace just blocks in a loop, so a timeout is needed.

fntlnz:
It seems like that the build fails when building in parallel.

```
ebpf-rhysre-libebpf-impl/build » make -j16
[ 29%] Built target BPFTcFilter
[ 44%] Built target gtest
[ 44%] Built target libelf
[ 46%] Built target libbpf
Consolidate compiler generated dependencies of target Common
Consolidate compiler generated dependencies of target BPFEventProbe
[ 48%] Built target BPFKprobeConnectHook
[ 50%] Building CXX object GPL/HostIsolation/TcFilter/CMakeFiles/BPFTcFilterTests.dir/BPFTcFilterTests.cpp.o
[ 51%] Building CXX object GPL/HostIsolation/TcFilter/CMakeFiles/BPFTcFilterTests.dir/__/__/__/contrib/googletest/googletest/src/gtest_main.cc.o
[ 53%] Built target BPFEventProbe
[ 57%] Built target Common
[ 59%] Building C object non-GPL/LibEbpfEvents/CMakeFiles/LibEbpfEvents.dir/LibEbpfEvents.c.o
[ 61%] Building C object non-GPL/TcLoader/CMakeFiles/TcLoader.dir/TcLoader.c.o
[ 62%] Building C object non-GPL/HostIsolation/KprobeConnectHook/CMakeFiles/KprobeLoader.dir/KprobeLoader.c.o
[ 64%] Building C object non-GPL/HostIsolationMapsUtil/CMakeFiles/UpdateMaps.dir/UpdateMaps.c.o
/home/fntlnz/Projects/elastic/ebpf-rhysre-libebpf-impl/non-GPL/LibEbpfEvents/LibEbpfEvents.c:11:10: fatal error: EventProbe.skel.h: No such file or directory
 #include "EventProbe.skel.h"
          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

non-GPL/LibEbpfEvents/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# SPDX-License-Identifier: LicenseRef-Elastic-License-2.0
+
+# Copyright 2021 Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+# or more contributor license agreements. Licensed under the Elastic License 2.0;
+# you may not use this file except in compliance with the Elastic License 2.0.
+
+add_library(LibEbpfEvents LibEbpfEvents.c)
+add_dependencies(LibEbpfEvents libbpf BPFEventProbe)
```
fntlnz, line 8:
```suggestion
add_dependencies(LibEbpfEvents EventProbe_skeleton libbpf BPFEventProbe)
```

fntlnz: ^ Fix the dependency chain to solve the parallel building problem.

GPL/Events/Helpers.h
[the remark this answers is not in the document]
fntlnz: Can we keep `FileEvents.h` and include it in `Helpers.h` instead?
In this way, we keep the same UX but things are already pretty splitted. IMO we don't need to defer this to a second moment, it's cheap to do it now.

non-GPL/LibEbpfEvents/CMakeLists.txt
rhysre: fixed

GPL/Events/Helpers.h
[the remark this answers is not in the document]
rhysre: Sounds good, done.

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -0,0 +1,92 @@
...
+    ebpf_event_handler_fn cb = ctx;
+    struct ebpf_event *evt = data;
+    return cb(evt, size);
+}
+
+int ebpf_event_ctx__new(
+        struct ebpf_event_ctx **ctx,
+        ebpf_event_handler_fn cb,
+        uint64_t features,
+        uint64_t events)
```
fntlnz, line 36: Event filtering seems to not be implemented but I don't suggest we should consider this as a blocker for this PR. I opened an issue for us to remember. 


-  https://github.com/elastic/ebpf/issues/49

fntlnz (approved):
LGTM, opened an issue as a follow up.

We can merge this and iterate more on the events.

Thanks @rhysre

GPL/EventProbe/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
+# SPDX-License-Identifier: GPL-2.0
```
leodido, line 1: Is this SPDX identifier correct, Rhys?

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -17,13 +17,12 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
+#define NULL 0
```
leodido, line 20: Shouldn't this come from the `bpf_helpers.h` file in the libbpf source?

https://github.com/libbpf/libbpf/blob/9fab7c81ecd462ce6aa1eabd24b230f0fc6a14c7/src/bpf_helpers.h#L57

Anyways, if you want to define it here, I would suggest to wrap it with `#ifndef`.

leodido (commented):
SGTM! Good job 💯 

Left two nits

GPL/EventProbe/CMakeLists.txt
rhysre: My understanding is that everything in `GPL/` (with the exception of EbpfEventProto.h) should be GPL'd, CMakeLists.txt or C source. All the other CMakeLists.txt were already GPL'd.

GPL/EventProbe/EventProbe.bpf.c
rhysre: That `#define` isn't available in the libbpf that currently exists under `contrib/`. We can remove it when we update libbpf to 0.6.0, but for now I'll just surround this with an ifdef. Good catch.

leodido: Ok, let's remember to remove when #48 lands

leodido: On a side note, Rhys:

While working on the kernel and other OSS projects, I learned that it’s a good practice to use the “Co-Authored-By” note in the body of the commit to attribute credit to people helping with suggestions/reviews.
I would really appreciate it if you can do this for my suggestions in this PR so that we can share ownership on that code and both be a reference when someone will need help with that code (or when we'll need to remove it because PR 48 landed).

Not to be pedantic, but it actually brings a few benefits:
- helps in understanding who to ping when there's a bug
- helps in attributing the work (important in OSS projects) 
- it's just a nice thing to do to show appreciation to the individuals helping you

My 2 cents :)

rhysre: Oh whoops, sorry. Alex brought this up with me but it totally slipped my mind. First time working in a commercial capacity in an open-source repo. Old habits die hard I guess 😆 

I'll add the `Co-Authored-By` tag to all future commits that need it. Thanks for the reminder Leo.

rhysre: Also #48 is now merged so I removed the define entirely.

fntlnz:
Anyone else can take a look ? This should be ready to go !

tabell (approved):

PR #52, opened 2021-12-28
Format code and enforce code formatting in Jenkins

### LINKED ISSUE #42: Choose a code style and define Clang format rules
Our clang format is currently empty and this leads to confusion.

We need to

- Pick a style for our C++ and C (Userspace and eBPF) code
- Write clang format rules, or use standard ones, to reflect that

Opening this because this was brought up many times by different team members so that we have a place to start a conversation.

rhysre:
This PR adds a `.clang-format`, formats the code, and updates the `Jenkinsfile` to error if code isn't formatted.

It's branched off #51 as it's not yet merged. Opening it up now to get reviews early.

Code can now be formatted as follows:

```bash
find GPL/ non-GPL/ -name "*.[c,h]" | xargs clang-format -i
```

Here's the `.clang-format` I added:

```yaml
BasedOnStyle: LLVM
IndentWidth: 4
BreakBeforeBraces: Linux
BinPackParameters: false
AlignConsecutiveAssignments: true
```

See all possible values that can go into a `.clang-format` [here](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

I'll give a small bit of rationale here for why I chose these values, but want to emphasize that I ultimately don't really care about the minutia of the code style, I just care that we have one and that it's applied consistently and enforced by Jenkins. The lack of consistent code formatting in this repo was really getting to me. This is code style, so all of the choices I've made are by definition subjective and up to the whims of personal preference (especially the last three). I'm not married to this `.clang-format`, happy to change it if everyone else is against me, I really just want consistent formatting.

Rationale for `.clang-format` values:

```yaml
BasedOnStyle: LLVM
```
Might as well choose something widely used for the base style. LLVM style isn't that esoteric (like GNU for example) and is pretty widely used.

```yaml
IndentWidth: 4
```
LLVM style specifies this as 2, but absolutely everyone here was using 4 already, so seems like we have consensus on that.

```yaml
BreakBeforeBraces: Linux
```
Uses the Linux brace breaking style, so function definitions have their opening `{` on a new line, but `if`/`while`/`for` keep it on the same line. IMO we should try to emulate kernel code style here since this is more-or-less kernel code. This is highly subjective however, happy to drop it if there are objections.

```yaml
BinPackParameters: false
```
Ensures that function prototypes all have their arguments on the same line, or, if they don't fit, each one will have one line each. e.g.

```c
true:
void f(int aaaaaaaaaaaaaaaaaaaa, int aaaaaaaaaaaaaaaaaaaa,
       int aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa) {}

false:
void f(int aaaaaaaaaaaaaaaaaaaa,
       int aaaaaaaaaaaaaaaaaaaa,
       int aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa) {}
```

IMO this is really nice for BPF probe definitions (which we will have a lot of), e.g.

```c
true:
int BPF_PROG(security_path_unlink_exit, const struct path *dir, 
                         struct dentry *dentry, long ret)

false:
int BPF_PROG(security_path_unlink_exit,
             const struct path *dir,
             struct dentry *dentry,
             long ret)
```

```yaml
AlignConsecutiveAssignments: true
```
Without this, enum definitions where each item has a specific value (which I anticipate a fair bit of), look wonky e.g.

```c
enum ebpf_event_type {
    EBPF_EVENT_PROCESS_FORK = (1 << 1),
    EBPF_EVENT_PROCESS_EXEC = (1 << 2),
    EBPF_EVENT_FILE_DELETE = (1 << 3),
};
```
vs:
```c
enum ebpf_event_type {
    EBPF_EVENT_PROCESS_FORK = (1 << 1),
    EBPF_EVENT_PROCESS_EXEC = (1 << 2),
    EBPF_EVENT_FILE_DELETE  = (1 << 3),
};
```

I'm also of the opinion that it enhances readability generally. It's also used in other C/C++ projects at Elastic.

Fixes #42

### DIFF
--- a/non-GPL/EventsTrace/EventsTrace.c
+++ b/non-GPL/EventsTrace/EventsTrace.c
@@ -23,8 +23,7 @@ static void sig_int(int signo)
     fprintf(stdout, "Received SIGINT, Exiting...\n");
 }
 
-static void ebpf_file_event_path__tostring(struct ebpf_file_path path,
-                                           char *pathbuf)
+static void ebpf_file_event_path__tostring(struct ebpf_file_path path, char *pathbuf)
 {
     strcpy(pathbuf, "/");
     for (int i = 0; i < path.patharray_len; i++) {
@@ -44,15 +43,9 @@ static void out_object_start() { printf("{"); }
 
 static void out_object_end() { printf("}"); }
 
-static void out_event_type(const char *type)
-{
-    printf("\"event_type\":\"%s\"", type);
-}
+static void out_event_type(const char *type) { printf("\"event_type\":\"%s\"", type); }
 
-static void out_int(const char *name, const int value)
-{
-    printf("\"%s\":%d", name, value);
-}
+static void out_int(const char *name, const int value) { printf("\"%s\":%d", name, value); }
 
 static void out_string(const char *name, const char *value)
 {
@@ -189,10 +182,9 @@ int main(int argc, char const *argv[])
     struct ebpf_event_ctx *ctx;
     uint64_t features = EBPF_KERNEL_FEATURE_BPF;
     uint64_t events   = EBPF_EVENT_FILE_DELETE;
-    err = ebpf_event_ctx__new(&ctx, event_ctx_callback, features, events);
+    err               = ebpf_event_ctx__new(&ctx, event_ctx_callback, features, events);
     if (err < 0) {
-        fprintf(stderr, "Could not create event context: %d %s\n", err,
-                strerror(-err));
+        fprintf(stderr, "Could not create event context: %d %s\n", err, strerror(-err));
         goto cleanup;
     }
 
--- a/non-GPL/EventsTrace/EventsTrace.c
+++ b/non-GPL/EventsTrace/EventsTrace.c
@@ -1,15 +1,16 @@
 // SPDX-License-Identifier: LicenseRef-Elastic-License-2.0
 
 /*
- * Copyright 2021 Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
- * or more contributor license agreements. Licensed under the Elastic License 2.0;
- * you may not use this file except in compliance with the Elastic License 2.0.
+ * Copyright 2021 Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under
+ * one or more contributor license agreements. Licensed under the Elastic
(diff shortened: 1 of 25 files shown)


stanek-michal:
Thanks for adding this. We definitely need enforced formatting. I agree with the starting parameters and I also agree that we should more or less follow kernel style for eBPF code. I looked at what options we've had in the Endpoint repo and I would consider adding these:
```
DerivePointerAlignment: false
PointerAlignment: Right
SpaceBeforeParens: ControlStatements
SpacesBeforeTrailingComments: 1
UseTab:          Never
```

Maybe a column limit as well (100-char?)

rhysre:
> Thanks for adding this. We definitely need enforced formatting. I agree with the starting parameters and I also agree that we should more or less follow kernel style for eBPF code. I looked at what options we've had in the Endpoint repo and I would consider adding these:
> 
> ```
> DerivePointerAlignment: false
> PointerAlignment: Right
> SpaceBeforeParens: ControlStatements
> SpacesBeforeTrailingComments: 1
> UseTab:          Never
> ```
> 
> Maybe a column limit as well (100-char?)

Thanks for the suggestions, these nicely bring us more in line with kernel style, added. Also added a 100 column limit.

non-GPL/EventsTrace/EventsTrace.c
```diff
@@ -25,46 +26,26 @@ static void sig_int(int signo)
...
             strcat(pathbuf, "/");
         }
     }
 }
 
-static void out_comma()
-{
-    printf(",");
-}
+static void out_comma() { printf(","); }
```
nicholasberlin, line 38: I'm not a fan of this style change, but perhaps it points out that these functions should be macros.

nicholasberlin (approved):

non-GPL/EventsTrace/EventsTrace.c
rhysre: Same here actually, this is un-kernellike, just realized it was doing this. 

Fixed with `AllowShortFunctionsOnASingleLine: false`

fntlnz (approved):
Thanks for making this happen Rhys!

PR #53, opened 2021-12-30
Add support for deleting a single item from map

stanek-michal:
Also make sure to report errors from bpf_map_get_next_key

### DIFF
--- a/non-GPL/HostIsolationMapsUtil/UpdateMaps.c
+++ b/non-GPL/HostIsolationMapsUtil/UpdateMaps.c
@@ -31,7 +31,10 @@ ebpf_create_map(enum ebpf_hostisolation_map map_id,
                 int *map_fd);
 static int
 ebpf_clear_map(const char *map_path,
-                enum ebpf_hostisolation_map map_id);
+               enum ebpf_hostisolation_map map_id);
+static int
+ebpf_map_delete_key(const char *map_path,
+                    const void *key);
 
 int
 ebpf_map_allowed_IPs_add(uint32_t IPaddr)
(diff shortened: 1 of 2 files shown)


ferullo (approved):
LGTM if it works.

This will be released in a patch release. Is the merge target ok for that?

stanek-michal:
> This will be released in a patch release. Is the merge target ok for that?

Not really, I was planning on merging this to `main` while actually checking out `delete_item_from_map_oldhead` in Endpoint, which is this single commit on top of the previous checked out Ebpf revision. This is because some additional work is needed to get Endpoint to work with `main` (modify build scripts, test if legacy kprobe loading still works on old kernels, switch to new toolchain in CI).

ferullo:
Hmm. Should we start a 7.16 branch of this repo then?

ferullo:
Thinking it more, we definitely should start a 7.16 branch based off the commit haha 7.16 has been released from in the past.

fntlnz (approved):
LGTM, thanks @stanek-michal

fntlnz:
Starting a 7.16 branch looks like a great idea, main is also under heavy development right now.

intxgo (approved):

non-GPL/HostIsolationMapsUtil/UpdateMaps.c
```diff
@@ -197,16 +269,27 @@ ebpf_clear_map(const char *map_path,
...
+        {
+            // map is already empty
+            rv = 0;
+            goto cleanup;
+        }
+        else
+        {
+            // failure (perhaps not supported)
+            rv = -1;
+            ebpf_log("Error getting next key while clearing map, errno=%d\n", errno);
```
ferullo, line 282: this line hits for me on a 4.15 kernel with errno 524

nicholasberlin: perhaps we need to clear `errno` before calling `bpf_map_get_next_key`?

nicholasberlin (approved):

non-GPL/HostIsolationMapsUtil/UpdateMaps.c
stanek-michal: I answered in the Endpoint PR which uses this - it turned out lpm_trie maps didn't have support for deleting keys until around 4.17.

Issue #46, opened 2021-12-16
File Events: Implement EBPF_EVENT_FILE_DELETE


fntlnz:
## Stories

- As a user of the `EventsTrace` program I want to be able to see printed on screen when I file is deleted.
- As a user of the libebpf library I want to be able to receive file deletion events when calling `ebpf_event__next`

## Data needed

<table>
<tr>
	<td>Field
	<td>Type
	<td>Description
<tr>
	<td>File Name
	<td>char[256]
	<td>New name of the file being deleted 
<tr>
	<td>Full File Path
	<td>char[4096]
	<td>Full source path of the file being deleted
</table>

## Probe

As a starting point , some possible hooking points for this event could be:

- ~fexit on `int security_path_unlink(const struct path *dir, struct dentry *dentry);` - Neds a kernel compiled with `CONFIG_SECURITY_PATH`~ - We can't do this, see this comment https://github.com/elastic/ebpf/issues/46#issuecomment-1000455585
- fexit on `iint vfs_unlink(struct user_namespace *mnt_userns, struct inode *dir,
	       struct dentry *dentry, struct inode **delegated_inode)`

## Action items

- [ ] Write one or more probes that can catch when a file is deleted
- [ ] Write BPF_PROG_TEST_RUN tests for the probes (if supported)
- [ ] Write integration tests for the probes
- [ ] Hook the event to be printed by the `EventsTrace` program

fntlnz:
An initial version of this program was added in https://github.com/elastic/ebpf/pull/40

We have a few issues to fix related to dentries (the following will likely impact this as well as #44  and #45) 

- [ ] Check if  the memory impact of `ebpf_event_file_path__from_dentry`  when building path arrays is lowe enough for the frequency of usage or needs to be optimized (initially suggested in review by @rhysre)
- [ ] Check if the path depth we are using is enough and what happens - What happens in such scenario needs to be investigated and we need to understand what are the limitations of this if any.
- [ ] When some conditions are met the root directory of the path is not found (@nicholasberlin was suggesting this might be because of the different mountpoints) - This needs to be investigaged and fixed

fntlnz:
The approach we are using right now with `security_path_unlink` needs to be changed in favor (probably) to `vfs_unlink` because AL2 do not have a kernel with `CONFIG_SECURITY_PATH` even though they ship very recent kernels. I suspect this might be common in other distros.

AL2 aarch64

```
[...]

PR #55, opened 2022-01-11
new(GPL/EventProbe): implement EBPF_EVENT_FILE_DELETE

fntlnz:
This is an alternative implementation to the `security_path_unlink` one
we were exploring until now. That implementation needs `CONFIG_SECURITY`
to work which is not present on most systems yet.

This also includes a mechanism to save the state of the current
file event between multiple programs using the tid.

This still has some issues to fix before this can be merged:
- [x] we need to instruct the dentry parameter for vfs unlink from userspace
  or via btf checks
- [x] there are some bugs with vfs mount (some paths are duplicated)

This also contains a few new macro to do relocations for kernel symbol arguments (they are not supported directly in libbpf).

Fixes https://github.com/elastic/ebpf/issues/46

Big thank you goes to @rhysre for sharing his experience on the the call chain and to @leodido @nicholasberlin  and @mmat11  for the ongoing help with the code and for discussing the relocation of function arguments .


fntlnz:
I rebased this branch against #57 and removed the dentry logic here. It looks like the bugs we had also went away with this new logic, I'll do some more testing to see if everything works as expected.

GPL/EventProbe/EbpfEventProto.h
```diff
@@ -54,7 +45,10 @@ struct ebpf_file_delete_event {
     struct ebpf_event_header hdr;
 
     struct ebpf_pid_info pids;
-    struct ebpf_file_path path;
+    // When computing the path we need to allocate twice the size of PATH_MAX
```
rhysre, line 48: If I'm not mistaken, this is wrong. The dentry resolver bounds all accesses to `path` by bounding the index to a maximum of `PATH_MAX` by doing `index & 4095`. A simple `PATH_MAX` is all we should need.

GPL/EventProbe/Maps.h
```diff
@@ -9,4 +9,39 @@ struct bpf_map_def SEC("maps") ringbuf = {
     .max_entries = 4096 * 64, // todo: Need to verify if 256 kb is what we want
 };
 
+enum ebpf_fileevents_tid_state_id {
```
rhysre, line 12: I'm not a huge fan of stuffing every map into `Maps.h`. The new maps you've added are basically an implementation detail of the file probes, and don't have to be included elsewhere. IMO it's cleaner to move them to the probe file.

If the probe.c file is getting cluttered, it's probably good to break it into different types of events. e.g. processprobe.c, fileprobe.c etc.

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -29,22 +29,69 @@
...
-SEC("fexit/security_path_unlink")
-int BPF_PROG(security_path_unlink_exit, const struct path *dir, struct dentry *dentry, long ret)
+// todo(fntlnz): this is a temporary workaround, the userspace program should do this
+// just for reference.
+// 2 is on 5.14 on arch
+// 1 is on 5.10 on al2
+// 1 is on 5.10.68 on COS
+const int fentry__vfs_unlink__dentry_parm = 2;
+
+SEC("fentry/do_unlinkat")
```
rhysre, line 39: When I wrote these probes, I was hooking every unlink syscall entry instead of just `do_unlinkat`. I like this better though. We don't have a "moving target problem" if a new unlink syscall is added. It'll just work if `unlinkat2` is added.

It does depend on `do_unlinkat` being the function that the kernel is calling for all unlink syscalls, which is subject to change, but I think it's overall better than hooking every syscall due to the aforementioned advantage.

rhysre (commented):
Some general comments.

GPL/EventProbe/EventProbe.bpf.c
fntlnz: Thanks for the context, I agree this seems like the right approach for now. In any case it is still very easy to change if we find any problems with it.

GPL/EventProbe/Maps.h
fntlnz: This reasoning makes sense to me, moving them.

GPL/EventProbe/EbpfEventProto.h
mmat11: I am working on a similar task and can confirm I get a verifier error by using anything less than `PATH_MAX*2-1`

rhysre: That could be because the variable you're using as a bound is spilled to the stack, or some other trickery (see here: https://github.com/elastic/ebpf/blob/main/GPL/EventProbe/PathResolver.h#L134).

The pathresolver code in #57 works just fine with a `PATH_MAX` sized-buffer by liberally applying bounds at every point where the verifier could be confused. What are you trying to read into the buffer and how are you doing it?

fntlnz: @rhysre From what I can see the cause of this cannot be attributed to the data being read.

This is caused by the fact that we are trying to write to contiguous memory (in a packed struct).

If the buffer we use, e.g: `path[PATH]` is not followed by contiguous memory that is at least twice the allocated memory for the buffer, the verifier will not think there's enough room to fit the copy.

```c
struct ebpf_file_delete_event {
    struct ebpf_event_header hdr;
    struct ebpf_pid_info pids;
    // When computing the path we need to allocate twice the size of PATH_MAX
    // because the verifier does not have a way to know if the path actually
    // fits in PATH_MAX
    char path[PATH_MAX];
} __attribute__((packed));
```

As you can see, we don't have anything in the struct after `path`.

In the case of #57 - the struct we copy into is:

```c
struct ebpf_process_exec_event {
    struct ebpf_event_header hdr;
    struct ebpf_pid_info pids;
    struct ebpf_cred_info creds;
    struct ebpf_tty_dev ctty;
    char filename[PATH_MAX];
    char cwd[PATH_MAX];
    char argv[ARGV_MAX];
} __attribute__((packed));
```

In this one we use the dentry resolver against `cwd` but the verifier will not complain because `argv` provides  enough room for the copy.

On `main`, try to move  `argv` right before `cwd` in that the `ebpf_process_exec_event` struct in your environment if you want to reproduce.

I'll debug the kernel a bit more to understand why this happens, if you have any ideas please lmk.

rhysre: Indeed, it works with exec due to the extra 8kb struct after `cwd`. I was wrong in my initial comment. This wasn't an issue we spotted at Cmd because we were using perfbuffers, and the allocated space we had for the event structs was very large to begin with (not compact like with ringbuf reservations).

The path resolver will never exceed a `PATH_MAX` sized buffer, there is appropriate bounds checking to ensure that. The verifier however isn't smart enough to see that. The bounds check looks like this:

```c
130         if (size + component.len + 1 > PATH_MAX) {
131             bpf_printk("path under construction is too long: %s", buf);
132             goto out_err;
133         }
```

We then call `bpf_probe_read_kernel_str` as follows:

```c
143         int ret = bpf_probe_read_kernel_str(buf + (size & PATH_MAX_INDEX_MASK),
144                                             PATH_MAX > size ? PATH_MAX - size : 0,
145                                             (void *)component.name);
```

`size` is the size of the path so far and `component.len` is the length of the current component we want to read. Looking at the verifier source, it seems like this math is too complicated for it to recognize that this check guarantees safety. We can prove to it that `size` is bounded or that `component.len` is bounded with a conditional or a logical and, but the problem rests with us having to prove to the verifier that _the sum of the two_ are bounded, which I don't think (looking at `kernel/bpf/verifier.h`) the verifier has any concept of.

I'm about 80% sure that given this verifier limit, the dentry walker can't be tweaked to support `PATH_MAX`-sized buffers at the end of a ringbuf reservation, so we're probably going to have to bump the size of the buffers to keep it happy. I didn't want to do that since it's ugly and leaks an implementation detail into the protocol, but I think it's the only way to solve this.

Do you think my reasoning is on point here? The verifier is a complicated beast and I could be wrong.

fntlnz: I have the same thoughts, I don't think this is something we can solve in our dentry walker. I'll debug a bit more tomorrow to understand exactly why this happens in the verifier to se if there's anything we can do.

Anyways, I'm still happy we have a solution, allocating more 8kb instead of 4 does not seem a bad penalty for what we do now.


While I was writing this, my brain just remembered that we had to do something similar in the Falco BPF probe for this exact reason https://github.com/falcosecurity/libs/blob/a75b361ca1df8713acc0636b943dd154144ea30d/driver/bpf/types.h#L164-L170

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -29,22 +29,103 @@
...
+    struct ebpf_fileevents_tid_state state;
+    __builtin_memset(&state, 0, sizeof(struct ebpf_fileevents_tid_state));
+    state.state_id = EBPF_FILEEVENTS_TID_STATE_UNLINK;
+    ebpf_fileevents_write_state__set(&state);
+
+    return 0;
+}
+
+SEC("fentry/mnt_want_write")
+int BPF_PROG(fexit__mnt_want_write)
```
mmat11, line 86: here we can use 
```suggestion
int BPF_PROG(fexit__mnt_want_write, struct vfsmount *m)
```
or you wanted to fill the map from userspace for all funcs/params even if nothing recently changed?

fntlnz: you are right, I wasn't able to see this changing in the kernels I played with. We can probably just use parameter expansion here and access context directly after if we have th eneed

fntlnz: done, thank you!

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -29,22 +29,99 @@
...
+    struct ebpf_fileevents_tid_state state;
+    __builtin_memset(&state, 0, sizeof(struct ebpf_fileevents_tid_state));
+    state.state_id = EBPF_FILEEVENTS_TID_STATE_UNLINK;
+    ebpf_fileevents_write_state__set(&state);
+
+    return 0;
+}
+
+SEC("fentry/mnt_want_write")
+int BPF_PROG(fexit__mnt_want_write, struct vfsmount *mnt)
```
nicholasberlin, line 83:
```suggestion
int BPF_PROG(fentry__mnt_want_write, struct vfsmount *mnt)
```
right?

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -29,22 +29,99 @@
...
     if (is_kernel_thread(task))
         goto out;
 
+    struct ebpf_fileevents_tid_state *state = ebpf_fileevents_write_state__get();
+    if (state == NULL) {
+        bpf_printk("vfs_unlink: no state\n");
+        goto out;
+    }
+
+    bpf_core_read(&de, sizeof(unsigned long long), ctx + fentry__vfs_unlink__dentry_parm);
```
nicholasberlin, line 110: Can you please explain why function parameter order is not handled by CO/RE infrastructure?

rhysre: There is no direct helper to detect the names/types of function arguments AFAIK (like `bpf_core_field_exists` but for function arguments). @fntlnz can we however use the `LINUX_KERNEL_VERSION` macro here to detect the function parameter index? I suspect we might not be able to due to distributions backporting the specific patch that changed the function signature to older kernels (we hit that problem at Cmd once).

fntlnz: @rhysre are you suggesting that we should compile this probe for every kernel with a compile macro just for this ?
The index you are observing here `fentry__vfs_unlink__dentry_parm` is calculated in user space from the BTF relocations (userspace code is not finished/pushed  yet) specifically to avoid this. 

This will be working regardless kernel versions and patches by leveraging (in user space l) the BTF function protos that are already available in the symbols.

rhysre: Oh sorry, I misspoke, it's not a macro, but a `__kconfig` extern. I mean doing something like this so it's a runtime operation:

```c
#include <bpf/bpf_helpers.h>

extern int LINUX_KERNEL_VERSION __kconfig;

...

if (LINUX_KERNEL_VERSION > KERNEL_VERSION(5, 15, 0)) {
    /* we are on v5.15+ */
}
```

That way we don't have to do any work in userspace at all.

This will however break if a distro uses a kernel that backports a patch that switched the order of the arguments but still reports an older kernel number. I know Ubuntu does this (e.g. Ubuntu may use a 5.8 kernel but backport a patch from 5.10 that switches the arg order, while still reporting 5.8 via the `LINUX_KERNEL_VERSION` macro). 

This caused us some problems at Cmd. I like your current method of using the BTF information to determine the index instead better due to this. It's far more robust. Clarifying what I mean regardless.

fntlnz: Ok I see what you mean, thanks for clarifying.

I did a lot of iterations on this and ultimately realized that the only reliable way to accomplish this is by looking at the relocations

I agree that the kernel version is not a reliable medium, I wouldn't even rely on `CONFIG_`  for the same reason.


I can probably spend a bit more time to make the ux of this better by hiding the complexity in an helper to avoid spilling those long ctx reads everywhere.

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -29,22 +29,99 @@
 
 char LICENSE[] SEC("license") = "GPL";
 
-SEC("fexit/security_path_unlink")
-int BPF_PROG(security_path_unlink_exit, const struct path *dir, struct dentry *dentry, long ret)
+enum ebpf_fileevents_tid_state_id {
```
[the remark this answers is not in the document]
fntlnz: I will move the map declarations and helpers into a specific file so we can keep the probes in this file  without everything else.

I did put this stuff in Maps initially, but Maps wasn't a good place either, finding a new home.

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -29,22 +29,99 @@
...
+    u64 tid = bpf_get_current_pid_tgid();
+    return bpf_map_update_elem(&elastic_ebpf_fileevents_tid_state, &tid, state, BPF_ANY);
+}
+
+// Context relocations: see the fill_ctx_relos in LibEbpfEvents.c
+// to see how these are updated from userspace
+const volatile int fentry__vfs_unlink__dentry_parm = 0;
+
+SEC("fentry/do_unlinkat")
+int BPF_PROG(fexit__do_unlinkat)
```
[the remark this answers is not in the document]
fntlnz: Absolutely, good catch Nick!

GPL/EventProbe/EventProbe.bpf.c
fntlnz: Thanks !!

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -29,22 +29,99 @@
...
     struct task_struct *task = bpf_get_current_task_btf();
     if (is_kernel_thread(task))
         goto out;
 
+    struct ebpf_fileevents_tid_state *state = ebpf_fileevents_write_state__get();
+    if (state == NULL) {
+        bpf_printk("vfs_unlink: no state\n");
+        goto out;
+    }
+
```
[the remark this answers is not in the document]
fntlnz: Unfortunately here we can't return anything other than zero so the only option I had was to log.

GPL/EventProbe/EventProbe.bpf.c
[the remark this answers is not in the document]
fntlnz: Done

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -38,12 +40,77 @@ static int ring_buf_cb(void *ctx, void *data, size_t size)
     return 0;
 }
 
+/* Find the BTF type relocation index for a named argument of a kernel function */
+static int resolve_btf_func_arg_id(const char *func_name, const char *arg_name)
```
rhysre, line 44: This is awesome 🚀

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -38,12 +40,77 @@ static int ring_buf_cb(void *ctx, void *data, size_t size)
...
+        goto out;
+
+    for (int i = 0; i < btf__type_cnt(btf); i++) {
+        int btf_type = btf__resolve_type(btf, i);
+        if (btf_type < 0)
+            continue;
+
+        const struct btf_type *btf_type_ptr = btf__type_by_id(btf, btf_type);
+
+        if (btf_is_func(btf_type_ptr)) {
```
rhysre, line 61: Instead of:

```c
if (condition) {
    /* lots and lots of code */
}
```

Can we do:

```c
if (!condition)
    continue;

/* Lots of code */
```

For the conditions on lines 61, 63 and 69.

This function is really heavily nested and hard to read as a result, flipping the conditionals would really help.

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -38,12 +40,77 @@ static int ring_buf_cb(void *ctx, void *data, size_t size)
...
+        }
+    }
+
+out:
+    return ret;
+}
+
+#define FILL_FUNCTION_RELO(ctx, func_name, arg_name)                                               \
+    ({                                                                                             \
+        int __r = 0;                                                                                   \
```
rhysre, line 89: Nit: `\` is indented slightly too much.

GPL/EventProbe/EventProbe.bpf.c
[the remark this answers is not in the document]
nicholasberlin: Oh, I mean, we should check the return value of `vfs_unlink` here, and if it's a failure, we shoulld jump to `out`. That way we don't emit a delete event when it wasn't successfully deleted.

non-GPL/LibEbpfEvents/LibEbpfEvents.c
fntlnz: Makes a lot of sense. Done

non-GPL/LibEbpfEvents/LibEbpfEvents.c
fntlnz: Done

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -10,7 +10,9 @@
 #define __aligned_u64 __u64 __attribute__((aligned(8)))
 #include "LibEbpfEvents.h"
 
-#include <bpf/libbpf.h>
+#include <bpf/bpf.h>
+#include <bpf/btf.h>
+#include <bpfok/libbpf.h>
```
mmat11, line 15:
```suggestion
#include <bpf/libbpf.h>
```

fntlnz: Thanks!

GPL/EventProbe/EventProbe.bpf.c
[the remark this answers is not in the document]
fntlnz: NICK!!! You are so right, completely missed this :D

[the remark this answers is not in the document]
fntlnz: Still trying to figure this out, since we can't use positional arguments we need to understand how to figure out the return value.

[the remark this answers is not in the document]
fntlnz: Done

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -38,14 +40,123 @@ static int ring_buf_cb(void *ctx, void *data, size_t size)
...
+    })
+
+/* Fill context relocations for kernel functions
+ * You can add additional functions here by using the FILL_FUNCTION_RELO macro
+ * Remember to declare it in `EventProbe.bpf.c` using the DECL_RELO_FUNC_ARGUMENT macro
+ */
+static int fill_ctx_relos(struct ebpf_event_ctx **ctx)
+{
+    int err = 0;
+    err     = FILL_FUNCTION_RELO(ctx, vfs_unlink, dentry);
```
mmat11, line 148: I think `err` needs to be checked every time

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -38,14 +40,123 @@ static int ring_buf_cb(void *ctx, void *data, size_t size)
...
+            goto out;
+        }
+    }
+
+out:
+    return ret;
+}
+
+/* Find the BTF relocation index for a func return value */
+static int resolve_btf_func_ret(const char *func_name)
```
mmat11, line 103: these two functions (`resolve_btf_func_arg_id` and `resolve_btf_func_ret`) feels very similar, what about keeping the old code and treat `ret` as a special arg? 

```
if ((strcmp(arg_name, "ret") == 0)
    ret = vlen
```

GPL/EventProbe/Helpers.h
```diff
@@ -22,13 +22,30 @@
...
+#define RELO_FENTRY_ARG_READ(type, func_name, arg_name)                                            \
+    ({                                                                                             \
+        type _ret;                                                                                 \
+        bpf_core_read(&_ret, sizeof(unsigned long long),                                           \
+                      ctx + arg__##func_name##__##arg_name##__);                                   \
+        _ret;                                                                                      \
+    })
+
+#define DECL_RELO_FUNC_RET(func_name) const volatile int ret__##func_name##__ = 0;
+#define RELO_FENTRY_RET_READ(type, func_name)                                                      \
```
mmat11, line 42: same here, we could skip these 2 macros if we treat `ret` as an arg

GPL/EventProbe/EventProbe.bpf.c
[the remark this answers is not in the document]
mmat11: (sorry for reopening :D) probably you forgot to check `ret`

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -38,14 +40,123 @@ static int ring_buf_cb(void *ctx, void *data, size_t size)
...
+        if ((*ctx)->probe->rodata->ret__##func_name##__ < 0)                                       \
+            __r = -1;                                                                              \
+        __r;                                                                                       \
+    })
+
+/* Fill context relocations for kernel functions
+ * You can add additional functions here by using the FILL_FUNCTION_RELO macro
+ * Remember to declare it in `EventProbe.bpf.c` using the DECL_RELO_FUNC_ARGUMENT macro
+ */
+static int fill_ctx_relos(struct ebpf_event_ctx **ctx)
```
mmat11, line 145: (reminder) as already discussed in chat this won't scale well with N funcs/args, it's ok for now but we should probably open a followup issue to optimize it

fntlnz: Can you elaborate a bit more on why this does not scale ? Do you think there’s a performance problem or a kind of discoverability problem ? This seems a pretty good solution for the need we have unless we have an alternative in mind.

the only alternative that comes on top of my mind is that we integrate this into libbpf directly to do it natively but that would really be a different thing.

non-GPL/LibEbpfEvents/LibEbpfEvents.c
fntlnz: I’ve been thinking about this but duplicating that code does not seem a big deal to me, I also don’t like the idea of “hiding” this implementation detail.

GPL/EventProbe/Helpers.h
fntlnz: As said above, this seems more semantically correct than treating the return as a magic argument to me.

non-GPL/LibEbpfEvents/LibEbpfEvents.c
fntlnz: makes sense, I forgot to do it

non-GPL/LibEbpfEvents/LibEbpfEvents.c
mmat11: correct me if I'm wrong, but this way we call N times this function (I don't see any caching/magic happening in libbpf) https://github.com/libbpf/libbpf/blob/3febb8a16597f4605450f6c947f4deefb446cb8a/src/btf.c#L1025 which is ok for now that we have few args to fill.

Maybe the optimal solution would be to have something (eg. a list of func/arg and fill rodata on the go)

PS. I tested locally and with 100 args it takes ~1second to startup, with 1k args ~8seconds so we can ignore this

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -38,14 +40,123 @@ static int ring_buf_cb(void *ctx, void *data, size_t size)
...
+    struct btf_param *params = btf_params(proto_btf_type_ptr);
+    for (int j = 0; j < btf_vlen(proto_btf_type_ptr); j++) {
+        const char *cur_name = btf__name_by_offset(btf, params[j].name_off);
+        if (strcmp(cur_name, arg_name) == 0) {
+            ret = j;
+            goto out;
+        }
+    }
+
+out:
```
mmat11, line 98: do we need `btf__free(btf)`?

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -38,14 +40,123 @@ static int ring_buf_cb(void *ctx, void *data, size_t size)
...
+
+    if (libbpf_get_error(btf))
+        goto out;
+
+    const struct btf_type *proto_btf_type_ptr = resolve_btf_type_by_func_name(btf, func_name);
+    if (!proto_btf_type_ptr)
+        goto out;
+
+    ret = btf_vlen(proto_btf_type_ptr);
+out:
```
mmat11, line 118: same here

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -23,28 +23,70 @@
...
+    struct ebpf_fileevents_unlink_state unlink_state;
+    unlink_state.mnt    = mnt;
+    state->state.unlink = unlink_state;
+    return 0;
+}
+
+SEC("fexit/vfs_unlink")
+int BPF_PROG(fexit__vfs_unlink)
 {
+    int ret                  = RELO_FENTRY_RET_READ(___type(ret), vfs_unlink);
```
leodido, line 63: I would check `ret` here and jump to `out` in case...

GPL/EventProbe/Helpers.h
```diff
@@ -22,13 +22,30 @@
...
 #include "EbpfEventProto.h"
-#include "FileEvents.h"
 
 #if BPF_DEBUG_TRACE == 0
 #undef bpf_printk
 #define bpf_printk(fmt, ...)
 #endif
 
+#define DECL_RELO_FUNC_ARGUMENT(func_name, arg_name)                                               \
+    const volatile int arg__##func_name##__##arg_name##__ = 0;
```
leodido, line 32: I'm glad we explained each other and you wrote this macro :)

Like it, as you already know I think it's the right approach

GPL/EventProbe/Helpers.h
```diff
@@ -22,13 +22,30 @@
...
+#define RELO_FENTRY_ARG_READ(type, func_name, arg_name)                                            \
+    ({                                                                                             \
+        type _ret;                                                                                 \
+        bpf_core_read(&_ret, sizeof(unsigned long long),                                           \
+                      ctx + arg__##func_name##__##arg_name##__);                                   \
+        _ret;                                                                                      \
+    })
+
+#define DECL_RELO_FUNC_RET(func_name) const volatile int ret__##func_name##__ = 0;
```
leodido, line 41: Maybe I recall it wrong but I remember `ret` to be `long` sometimes... Not a big issue IMO, but worth double-checking, WDYT :)

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -23,28 +23,70 @@
...
+    ebpf_fileevents_write_state__set(&state);
+    return 0;
+}
+
+SEC("fentry/mnt_want_write")
+int BPF_PROG(fentry__mnt_want_write, struct vfsmount *mnt)
+{
+    struct ebpf_fileevents_tid_state *state = ebpf_fileevents_write_state__get();
+    if (state == NULL) {
+        return 0;
```
rhysre, line 51: Can we `bpf_printk` an error message in all such error cases like this before exit? 

`bpf_printk` is disabled in non-debug builds now (so no performance penalty) and this will allow for very quickly finding the root of the problem if a probe is faulty.

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -23,28 +23,70 @@
...
+    }
+
+    struct ebpf_fileevents_unlink_state unlink_state;
+    unlink_state.mnt    = mnt;
+    state->state.unlink = unlink_state;
+    return 0;
```
rhysre, line 57: Some probes are now using `return 0` everywhere and some are using `goto out` / `out: return 0`. Can we make it consistent?

GPL/EventProbe/FileEventsHelpers.h
```diff
@@ -0,0 +1,58 @@
...
+#define EBPF_EVENTPROBE_FILEEVENTS_HELPERS_H
+
+enum ebpf_fileevents_tid_state_id {
+    EBPF_FILEEVENTS_TID_STATE_UNKNOWN = 0,
+    EBPF_FILEEVENTS_TID_STATE_UNLINK  = 1,
+};
+
+struct ebpf_fileevents_unlink_state {
+    struct vfsmount *mnt;
+};
```
rhysre, line 31: nit: newline needed between struct definitions.

GPL/EventProbe/EventProbe.bpf.c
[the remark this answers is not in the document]
rhysre: ^ What Matt said :P

leodido (commented):
Just some nits

It LGTM and the functions to find the params index is just 🔥

rhysre (commented):
Few more comments

GPL/EventProbe/EventProbe.bpf.c
nicholasberlin: +1

GPL/EventProbe/FileEventsHelpers.h
```diff
@@ -0,0 +1,58 @@
...
+static __always_inline struct ebpf_fileevents_tid_state *ebpf_fileevents_write_state__get(void)
+{
+    u64 tid = bpf_get_current_pid_tgid();
+    return bpf_map_lookup_elem(&elastic_ebpf_fileevents_tid_state, &tid);
+}
+
+static __always_inline long
+ebpf_fileevents_write_state__set(struct ebpf_fileevents_tid_state *state)
+{
+    u64 tid = bpf_get_current_pid_tgid();
```
nicholasberlin, line 55: Oh, I see. I thought this was a bug at first.

Instead of doing the shifting dance to isolate the pid, you are just using the combo of pid and tgid, which works too.

I was confused because of the variable and structure name has `tid` in them.

GPL/EventProbe/Helpers.h
```diff
@@ -22,13 +22,30 @@
...
+#define RELO_FENTRY_RET_READ(type, func_name)                                                      \
+    ({                                                                                             \
+        type _ret;                                                                                 \
+        bpf_core_read(&_ret, sizeof(unsigned long long), ctx + ret__##func_name##__);              \
```
nicholasberlin, line 45:
```suggestion
        bpf_core_read(&_ret, sizeof(_ret), ctx + ret__##func_name##__);              \
```
right? when `_ret` is an `int` we would be reading 64bits (`long long`) and storing in 32bits (`int`) right?

GPL/EventProbe/Helpers.h
```diff
@@ -22,13 +22,30 @@
...
 #undef bpf_printk
 #define bpf_printk(fmt, ...)
 #endif
 
+#define DECL_RELO_FUNC_ARGUMENT(func_name, arg_name)                                               \
+    const volatile int arg__##func_name##__##arg_name##__ = 0;
+#define RELO_FENTRY_ARG_READ(type, func_name, arg_name)                                            \
+    ({                                                                                             \
+        type _ret;                                                                                 \
+        bpf_core_read(&_ret, sizeof(unsigned long long),                                           \
```
nicholasberlin, line 36:
```suggestion
        bpf_core_read(&_ret, sizeof(_ret),                                           \
```

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -38,14 +40,123 @@ static int ring_buf_cb(void *ctx, void *data, size_t size)
...
+        if (btf_type < 0)
+            continue;
+
+        const struct btf_type *btf_type_ptr = btf__type_by_id(btf, btf_type);
+
+        if (!btf_is_func(btf_type_ptr))
+            continue;
+
+        const char *name = btf__name_by_offset(btf, btf_type_ptr->name_off);
+        if (!strcmp(name, func_name) == 0)
```
nicholasberlin, line 56:
```suggestion
        if (strcmp(name, func_name))
```
nit pick.

GPL/EventProbe/Helpers.h
```diff
@@ -22,13 +22,30 @@
...
+                      ctx + arg__##func_name##__##arg_name##__);                                   \
```
nicholasberlin, line 37: The arguments in BTF are zero indexed, right? (just making sure)

GPL/EventProbe/Helpers.h
fntlnz: you are correct here, now that we use `type` we can switch, we used `void *` before (but that wasn't compatible with scalar arguments). Nice find!

GPL/EventProbe/Helpers.h
fntlnz: Same answer as the other _ARG macro, thanks!

non-GPL/LibEbpfEvents/LibEbpfEvents.c
fntlnz: you're right, I was just lazy

GPL/EventProbe/Helpers.h
fntlnz: Yes they are

non-GPL/LibEbpfEvents/LibEbpfEvents.c
fntlnz: yes thank you!

fntlnz: yes thank you!

GPL/EventProbe/Helpers.h
fntlnz: This is always int because this is represents the index where to find the return value inside `ctx , not the value itself.

GPL/EventProbe/Helpers.h
fntlnz: done, thank you!

GPL/EventProbe/Helpers.h
fntlnz: done, thank you!

GPL/EventProbe/FileEventsHelpers.h
fntlnz: done, thank you!

GPL/EventProbe/EventProbe.bpf.c
fntlnz: done, thank you!

GPL/EventProbe/EventProbe.bpf.c
fntlnz: done, thank you!

GPL/EventProbe/Helpers.h
fntlnz: context for everyone: Leo suggested to use rodata instead of an hand crafted map, we authored an implementation in [`d6e5d57` (#55)](https://github.com/elastic/ebpf/pull/55/commits/d6e5d57f447fcf4a6361f6ec5b8ac20618b4da84) before I converted the definitions to macros.

GPL/EventProbe/EventProbe.bpf.c
fntlnz: done, thank you!

non-GPL/LibEbpfEvents/LibEbpfEvents.c
fntlnz: done, thank you!

non-GPL/LibEbpfEvents/LibEbpfEvents.c
fntlnz: done, thank you!

non-GPL/LibEbpfEvents/LibEbpfEvents.c
fntlnz: done, thank you!

GPL/EventProbe/EventProbe.bpf.c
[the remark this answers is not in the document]
fntlnz: done, thank you!

fntlnz:
All the comments are addressed, @leodido @rhysre @mmat11 @nicholasberlin - Let's merge this?

nicholasberlin (approved):

mmat11 (approved):

rhysre (approved):

PR #58, opened 2022-01-12
Disambiguate licensing to GPL 2.0 only

rhysre:
There were two problems with the license headers as things stood:

- The SPDX license identifiers used the deprecated "GPL-2.0" instead of
  the newer and less ambigous "GPL-2.0-only"
- The longer GPL header comments all stated that the files were licensed
  under GPL v3 or later, while the SPDX identifier above said GPL 2.0

This commit changes all the SPDX "GPL-2.0" headers to "GPL-2.0-only" and
removes the longer GPL header comments entirely. It seems the FSF
doesn't have a version that states GPL 2/3 _only_ and the license is
stated in the SPDX identifier (and LICENSE file ) anyways, so it's
redundant.

CC: @stanek-michal you last changed licensing in a major way (dbca96c1173d09bee9ffca915f8f7bcdc0bad40b) so want your sign-off on this before I merge.

This PR also updates the `LicenseRef-Elastic-License-2.0` SPDX identifiers to simply `Elastic-2.0`, since it's now officially part of the SPDX spec.

### DIFF
--- a/GPL/EventProbe/EbpfEventProto.h
+++ b/GPL/EventProbe/EbpfEventProto.h
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: GPL-2.0-only OR LicenseRef-Elastic-License-2.0
+// SPDX-License-Identifier: GPL-2.0-only OR Elastic-2.0
 
 /*
  * Copyright 2021 Elasticsearch BV
--- a/GPL/EventProbe/EbpfEventProto.h
+++ b/GPL/EventProbe/EbpfEventProto.h
@@ -1,6 +1,8 @@
-// SPDX-License-Identifier: GPL-2.0 OR LicenseRef-Elastic-License-2.0
+// SPDX-License-Identifier: GPL-2.0-only OR LicenseRef-Elastic-License-2.0
 
 /*
+ * Copyright 2021 Elasticsearch BV
+ *
  * This file is dual-licensed under the GNU General Public License version 2
  * and the Elastic License 2.0. You may choose either one of them if you use
  * this file.
--- a/GPL/EventProbe/BPFEventsTests.cpp
+++ b/GPL/EventProbe/BPFEventsTests.cpp
@@ -1,21 +1,20 @@
-// SPDX-License-Identifier: GPL-2.0
+// SPDX-License-Identifier: GPL-2.0-only
 
 /*
- * Elastic eBPF
- * Copyright 2021 Elasticsearch BV
+ * Copyright (C) 2021 Elasticsearch BV
  *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation, either version 3 of the License, or
- * (at your option) any later version.
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the Free
+ * Software Foundation; version 2.
  *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+ * more details.
  *
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, write to the Free Software Foundation, Inc., 51
+ * Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
  */
 
 #include <sys/resource.h>
--- a/GPL/EventProbe/CMakeLists.txt
+++ b/GPL/EventProbe/CMakeLists.txt
@@ -1,22 +1,19 @@
-# SPDX-License-Identifier: GPL-2.0
+# SPDX-License-Identifier: GPL-2.0-only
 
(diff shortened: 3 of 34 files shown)


stanek-michal:
As for the switch from GPL-2.0 to GPL-2.0-only, I think it's fine. But I think we probably still want a copyright notice in each file, which is a separate thing from the license. Perhaps we can make it a single line below the SPDX identifier? Something like:
```
// SPDX-License-Identifier: GPL-2.0-only
// Copyright 2021 Elasticsearch BV
```

rhysre:
> As for the switch from GPL-2.0 to GPL-2.0-only, I think it's fine. But I think we probably still want a copyright notice in each file, which is a separate thing from the license. Perhaps we can make it a single line below the SPDX identifier? Something like:
> 
> ```
> // SPDX-License-Identifier: GPL-2.0-only
> // Copyright 2021 Elasticsearch BV
> ```
> 
> or
> 
> ```
> // SPDX-License-Identifier: GPL-2.0-only
> 
> // Copyright 2021 Elasticsearch BV
> ```

Sounds good, thanks for the pointer Michal. 

Added short copyright notices back to all the files I stripped the GPLified copyright notice from.

stanek-michal (approved):
Looks good to me. Adding @ferullo as he was also involved in the initial license disclaimer

ferullo (dismissed):
Why is the bulk of the GPL text removed from each file?

stanek-michal:
> Why is the bulk of the GPL text removed from each file?

Rhys mentioned that the GNU website didn't have a  GPL-2-only variant of the license disclaimer text.
I was wondering if a single SPDX line was enough and I found this article on LWN: 

https://lwn.net/Articles/739183/
> An additional goal is to eventually get rid of the other license texts; the consensus seems to be that the SPDX identifier is a sufficient declaration of the license on its own.

The article mentions that there are a lot of variations of that license disclaimer and that a unified SPDX identifier solves the confusion. 

This [article](https://www.gnu.org/licenses/gpl-howto.en.html) on gnu.org mentions the license notice but it doesn't say anything about SPDX so it was probably written before that.

No strong opinions here though, not sure if Legal had any specific requirements about the GPL notices.

GPL/EventProbe/EbpfEventProto.h
```diff
@@ -1,4 +1,5 @@
-// SPDX-License-Identifier: GPL-2.0 OR LicenseRef-Elastic-License-2.0
+// SPDX-License-Identifier: GPL-2.0-only OR LicenseRef-Elastic-License-2.0
```
nicholasberlin, line 1: Seems strange to me that something could be GPL-2.0-*only* and something else. 🤷

gtback:
> @gtback do you have any insight into how the GPL headers should be marked?

@ferullo @rhysre I'll double-check in our Open Source Working Group meeting, but I am hopeful that just the SPDX tag will be sufficient.

If with go with the SPDX format, we can consider using `SPDX-FileCopyrightText: ` for the copyright statement ([SPDX standard](https://spdx.github.io/spdx-spec/file-tags/#h2-format)) to make this data more machine-parseable going forward.

I likewise was unable to find a "GPL-2.0-only" file header, but will look around more to see what other projects are doing.

GPL/EventProbe/EbpfEventProto.h
gtback: > LicenseRef-Elastic-License-2.0
 
@rhysre This is fine for now, but as of the next SPDX release, we [have an approved SPDX Identifier](https://github.com/spdx/license-list-XML/issues/1349): `Elastic-2.0` (and without the `LicenseRef` that indicates a "custom" license).

gtback:
> I likewise was unable to find a "GPL-2.0-only" file header, but will look around more to see what other projects are doing.

Of course, as soon as I said this I looked at the SPDX site and found this (under "Standard License Header" at the bottom):

> Copyright (C) yyyy name of author
> 
> This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; version 2.
> 
> This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
> 
> You should have received a copy of the GNU General Public License along with this program; if not, write to the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.

rhysre:
> > I likewise was unable to find a "GPL-2.0-only" file header, but will look around more to see what other projects are doing.
> 
> Of course, as soon as I said this I looked at the SPDX site and found this (under "Standard License Header" at the bottom):
> 
> > Copyright (C) yyyy name of author
> > This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; version 2.
> > This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
> > You should have received a copy of the GNU General Public License along with this program; if not, write to the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.

Oh fantastic, didn't see that on the SPDX website because it was hiding at the bottom. I added back the long-form GPL headers that say GPLv2 only so we can keep the existing style and just correct the v2/v3 error. Thanks for pointing that out Greg.

fntlnz (approved):
LGTM but needs a rebase. Thanks for looking into this @rhysre

rhysre:
> LGTM but needs a rebase. Thanks for looking into this @rhysre

Done. I also pushed another commit changing all `LicenseRef-Elastic-License-2.0` SPDX headers to simply `Elastic-2.0` since [it's now officially a part of the SPDX spec](https://spdx.org/licenses/Elastic-2.0.html) as @gtback stated above.

@ferullo Can I get a +1 for you to unblock merge?

PR #64, opened 2022-01-25
Two small BPF fixups

rhysre:
- Fix `argv` array potentially containing garbage on exec
- Don't track thread start/stop if it's not the thread group leader

Thanks for the catches on both counts @m-sample


GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -104,7 +104,7 @@ int BPF_PROG(sched_process_exec,
 SEC("tp_btf/sched_process_exit")
 int BPF_PROG(sched_process_exit, const struct task_struct *task)
 {
-    if (is_kernel_thread(task))
+    if (!is_thread_group_leader(task) || is_kernel_thread(task))
```
ferullo, line 107: The main thread of execution in a process can exit even while threads keep running. What would happen here if that's the case? Would this still wait to report when the full process exits (in other words is the thread group leader automatically reassigned?) or would a termination event be emitted too soon?

```
vagrant@vagrant:~/endpoint-dev$ cat example.c 
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <syscall.h>
#include <time.h>
#include <unistd.h>

void*
thread_entry(
    void *unused)
{
    time_t now = time(NULL);

    printf("hello from thread %s", ctime(&now));

    sleep(10);

    now = time(NULL);
    printf("goodbye from thread %s", ctime(&now));
}

int main(
int argc,
char *arg[])
{
    pthread_t thread;
    time_t now = time(NULL);

    printf("hello from main %s", ctime(&now));

    (void)pthread_create(&thread, NULL, thread_entry, NULL);

    /* exit just this thread */
     syscall(__NR_exit);

    now = time(NULL);
    printf("goodbye from main %s", ctime(&now));

    return 0;
}
vagrant@vagrant:~/endpoint-dev$ gcc -o example example.c -lpthread; ./example
hello from main Tue Jan 25 02:09:49 2022
hello from thread Tue Jan 25 02:09:49 2022
goodbye from thread Tue Jan 25 02:09:59 2022
vagrant@vagrant:~/endpoint-dev$ 
```

nicholasberlin: perhaps we can filter on `current->signal->live` being zero?
https://elixir.bootlin.com/linux/v5.16.2/source/kernel/exit.c#L794

ferullo: It's been a long time, but I think that if the two final threads exit at the same time then that could lead to emitting two exit events as a race condition. Event filtering could mitigate that of course, if my memory is right in the first place.

fntlnz: I think that @nicholasberlin solution is the way to go. We need to determine if the current thread is the last one in the group.

`signal->live` gets to zero only in that case from what I can see.

lrishi: From what I understand, we're trying to ignore thread exits. In this case, we'll also need to ensure that ebpf_pid_info__fill() uses tgid as pid (which I assume is the event key) to submit the exit event, else we're inconsistent.

nicholasberlin: `current->signal->nr_threads` seems like a good option as well.

m-sample: My observation is the same as Dan's - if the pid==tgid thread exits before some other threads in the process, you will get TWO exit events for the pid=tgid thread/process.

This can be reproduced by using `go version` on Ubuntu if Go is installed from snap.  (snap use a busy-box like symlink'd exe layout with lots of execs within a given process, ostensibly for security sandboxing).  You will see two exits for the 'go version' process every time you run it

I'm guessing access to "nr_threads" and "live" may have race conditions / require a lock

Worst case, if we can't filter the first exit of the two, user space could update the data model with the last exit.  Protocol-wise it's a bit harder to delay sending an exit event because one my come in an indeterminate time in the future - so it would really good to solve this.

lrishi: @m-sample Tried this with bpftrace, I see main thread exits first, and then we get 4 more exit events, one for each thread with pid set to the thread id.

# sudo bpftrace -e 'tracepoint:sched:sched_process_exit {printf("%s exit: [pid(kernel tgid)=%d,tid(kernel pid)=%d]\n", comm, pid, tid);}'
Attaching 1 probe...
go exit: [pid(kernel tgid)=3801,tid(kernel pid)=3801]
go exit: [pid(kernel tgid)=3801,tid(kernel pid)=3802]
go exit: [pid(kernel tgid)=3801,tid(kernel pid)=3805]
go exit: [pid(kernel tgid)=3801,tid(kernel pid)=3804]
go exit: [pid(kernel tgid)=3801,tid(kernel pid)=3806]
go exit: [pid(kernel tgid)=3801,tid(kernel pid)=3803]

Not sure if bpftrace is interpreting the info differently

lrishi: ah, got it, we use tgid as pid in our application.

rhysre: The most educational days at work are those where I spend 8 hours working on a three line patch. This is one of those days.

There are two distinct issues at play here that have been raised in this thread and confused a bit.

### Issue 1

Raised by @ferullo in the first comment. The thread group leader is not reassigned, so yes, this PR as it currently stands will emit a termination event too soon.

We can use `current->signal->live` to check if the exiting thread is the last one in the process group, however, as mentioned, this seems prone to race conditions. Looking at `kernel/exit.c` for a 5.15.2 kernel, `current->signal->live` is decremented in `do_exit` on [line 794](https://elixir.bootlin.com/linux/v5.16.2/source/kernel/exit.c#L794) and the `sched_process_exit` tracepoint is invoked on [line 823](https://elixir.bootlin.com/linux/v5.16.2/source/kernel/exit.c#L794). The space between these lines is not a critical section protected by a lock (`current->signal->live` seems to be an atomic for exactly that reason) meaning that poor timing could lead to a duplicate event if two threads exit on different cores in very short order.

As raised by @nicholasberlin, `current->signal->nr_tasks` shows some promise, but is subtly different from `live`. `nr_tasks` measures the number of _allocated_ task structs for a given thread group. This includes threads that have exited and are zombies, meaning that if we use it instead we wouldn't emit an exit event if there were any zombies hanging around in the thread group.

A somewhat ugly but working way of getting around this race would be to use a kprobe on `taskstats_exit` instead of a the `sched_process_exit` tracepoint. [`taskstats_exit` is only called from `do_exit`](https://elixir.bootlin.com/linux/v5.16.2/source/kernel/exit.c#L817) and is passed a variable `group_dead`, the result of atomically decrementing `current->signal->live` and testing if it's equal to 0. We could only emit an exit event if that kprobe was invoked with the `group_dead` argument being equal to true (which will only happen once thanks to the atomic decrement). This _should_ solve the race condition, but would mean using a kprobe over the unchanging API of the tracepoint, which is somewhat ugly.

What are everyone's thoughts on the best solution to this? Please poke holes in my logic if I've gotten something wrong. Might be worth mentioning that the Cmd probes use the same race-y logic that's currently in this PR in the process exit probe and to my knowledge, we've never noticed it. Perhaps we want to fix it over there too?

### Issue 2

This one is more insidious but we're arguably doing the correct thing right now.

Raised by @m-sample. The case where _two_ exit events are emitted is distinct and unrelated to issue 1. It does not happen when the thread group leader exits while there are still other running threads. Rather, it happens when a _non-thread-group-leader_ `exec`'s something _after_ its thread group leader has exited (this is what snap is doing under the hood when you run `go version`). Example code to trigger it:

```c
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <syscall.h>
#include <time.h>
#include <unistd.h>

void*
thread_entry(void *unused)
{
    sleep(2);
    execl("/bin/ls", "/bin/ls");

    return NULL;
}

int main(int argc, char *arg[])
{
    pthread_t thread;
    time_t now = time(NULL);

    printf("thread group leader has pid %d", getpid());

    (void)pthread_create(&thread, NULL, thread_entry, NULL);

    /* exit just this thread */
     syscall(__NR_exit);

    return 0;
}
```

When the thread performs an `execve`, all other threads in the thread group are terminated, and the thread performing the exec [_assumes the pid_ of the thread group leader](https://elixir.bootlin.com/linux/v5.16.2/source/fs/exec.c#L1073). This means that we can have the following course of events:

1. Process starts as (tgid 1000 tid 1000)
2. Process `clone`'s a thread as (tgid 1000 tid 1001)
3. (tgid 1000, tid 1000) exits, exit event is emitted with tgid 1000, tid 1000
4. (tgid 1000, tid 1001) does `execve("/bin/ls")` and _assumes new pid info of_ (tgid 1000, tid 1000)
5. /bin/ls exits, duplicate exit event is emitted with tgid 1000, tid 1000

The behaviour the probes are currently exhibiting (emitting a duplicate event for tgid/tid 1000/1000) is arguably correct. A process with that tid _does_ exit twice, regardless of how confusing that output may be. This issue is probably outside the scope of this PR and should be discussed separately. I'm guessing the duplicate exec may present a corner case when it comes to nicely rendering this info in a webapp, which we should probably think about.

lrishi: @rhysre great summary, and really appreciate the pointers, they're always a good read. I do believe the 2nd case you're referring to is a non-issue, since we have always handled an exec followed by exit. This isn't exactly a duplicate exit event, its just that we miss a fork event, which is already handled (or should be) by our code, given that we're not the first process to start and we'll always have missing fork events.
The sequence here would be fork(1000) -> exit(1000), new-process,same pid(missing fork) -> exec(1000) -> exit(1000), which is quite expected, and at no point will you get an exit event with pid=1000, tid=1001

The original discussion is about a case which the code at the end will create:
test exit: [pid(kernel tgid)=2588,tid(kernel pid)=2588]
test exit: [pid(kernel tgid)=2588,tid(kernel pid)=2589]
here, you get two exit events for same tgid.

What @m-sample and @ferullo suggested is accurate that, we should be able to ignore the 2nd event safely, even if it is arbitrarily in future.
@ferullo I do see endpoint running checks like pProcessEvent->Tid() == pid at multiple places, to handle such cases, but not sure if its exhaustive, given the new data, need to dig deeper.

Try this:
`#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <syscall.h>
#include <time.h>
#include <unistd.h>
void* thread_entry(void *unused) {
    sleep(2);
    //execl("/bin/ls", "/bin/ls");
    return NULL;
}
int main(int argc, char *arg[]) {
     pthread_t thread;

     time_t now = time(NULL);
     printf("thread group leader has pid %d", getpid());
     (void)pthread_create(&thread, NULL, thread_entry, NULL);
     /* exit just this thread */ syscall(__NR_exit);
     return 0;
}
`

fntlnz: I agree with @rhysre when he says we should use a kprobe (or a  fentry) in a place where we know for sure the group is exited.

m-sample: > What are everyone's thoughts on the best solution to this?

If the taskstats_exit function has been stable for quite a few kernel releases, it seems like a good place to capture this information without races.  Of course we'll have to watch new kernel releases carefully for breaking changes.

Nice work on getting to the bottom of the double exit events for the same process.  This may be a common situation with snaps?  I think we'll need to send both exit events (with their different timestamps) to the cloud to ensure all potential rendering options are correct (i.e. they do not show a process as exited and then subsequently exec'ing).

rhysre: Taking a look at the `git blame` for `kernel/exit.c`, the call to `taskstats_exit` (and the function itself) has been there since 2006 (torvalds/linux@115085ea0794c0f339be8f9d25505c7f9861d824), so we're good as far as stability goes. We will of course, have to monitor the `do_exit` function in the kernel in case it is ever changed in the future.

This patch fixes issue 1 in a race-condition-free manner by using a `fentry` on `taskstats_exit`:

```diff
diff --git a/GPL/EventProbe/EventProbe.bpf.c b/GPL/EventProbe/EventProbe.bpf.c
index de62244..56be676 100644
--- a/GPL/EventProbe/EventProbe.bpf.c
+++ b/GPL/EventProbe/EventProbe.bpf.c
@@ -101,10 +101,12 @@ out:
     return 0;
 }

-SEC("tp_btf/sched_process_exit")
-int BPF_PROG(sched_process_exit, const struct task_struct *task)
+SEC("fentry/taskstats_exit")
+int BPF_PROG(fentry__taskstats_exit,
+        const struct task_struct *task,
+        int group_dead)
 {
-    if (!is_thread_group_leader(task) || is_kernel_thread(task))
+    if (!group_dead || is_kernel_thread(task))
         goto out;

     struct ebpf_process_exit_event *event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
```

@lrishi and I just had a productive conversation where he brought up that the patch above also changes the behaviour with regards to issue 2. In fixing issue 1 with the above patch, we will be causing the first exit event in issue 2 to not be emitted. If you think about it, the first exit event emitted in issue 2 *is in-and-of-itself* issue 1 (an exit event being emitted while there are still threads running in the thread group).

@m-sample Given that, it seems cleaner to use the above patch and only emit an exit upon every thread in a thread group exiting, regardless of whether or not the leader of that thread group has been "taken over" by an exec'ing thread. Even if the thread group leader has exited and been subsequently "taken over", that thread group is still, and has always been running, thus it seems to me that an exit should not be emitted. Does that seem reasonable to you? Intuitively, I wouldn't expect this to cause many problems for the webapp, an exit event will only be emitted when every thread is dead, thus no more events can come after an exit in any circumstances.

Also discussed with @lrishi was the fact that there are a lot of `if (Tid() == Tgid()) { drop_event(); }` checks in the current endpoint codebase to filter out events from non-thread-group-leaders. Since execs/forks/exits may be performed by non-thread-group-leaders, we should be very careful that our userspace enrichment path doesn't accidentally filter-out execs/forks if they happen to be performed by a non-thread-group-leader. Before shipping, we should absolutely have an integration test hitting the case where e.g. a thread group exit is performed by a non-thread-group-leader or a non-thread-group-leader performs a `fork()` to ensure that data doesn't get dropped on the enrichment path and makes it to Elasticsearch.

m-sample: is  "if (Tid() != Tgid()) { drop_event; }" what you intended to type?

The rest sounds great - exactly one exit event when all threads are gone/the process is truly done.  Agree that sched_process_fork and sched_process_exec probes should not filter out those from non-thread-group leaders - any thread can fork or exec.

Agree on testing - these nice little C examples in this review ideally make it to the test suite

rhysre: > is "if (Tid() != Tgid()) { drop_event; }" what you intended to type?
 
Yep, sorry typo there on my part, I meant `Tid() != Tgid()`.

I just force-pushed a fix that takes into account everything discussed in this thread and rebased on main.

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -146,10 +159,30 @@ int BPF_PROG(sched_process_exec,
...
 
-SEC("tp_btf/sched_process_exit")
-int BPF_PROG(sched_process_exit, const struct task_struct *task)
+// Process exit probe
+//
+// Note that we aren't using the sched_process_exit tracepoint here as it's
+// prone to race conditions. We want to emit an exit event when every single
+// thread in a thread group has exited. If we were to try to detect that by
+// checking task->signal->live == 0 (i.e. check that there are now 0 running
+// thread in the thread group), we would race with a thread exit on another CPU
```
nicholasberlin, line 168:
```suggestion
// threads in the thread group), we would race with a thread exit on another CPU
```

nicholasberlin (approved):
Nice comments, the future thanks you.

rhysre:
Bump. Can we get another +1 on this for merge?

fntlnz (approved):

tabell (approved):

Issue #44, opened 2021-12-16
File Events: Implement EBPF_EVENT_FILE_RENAME


fntlnz:
## Stories

- As a user of the `EventsTrace` program I want to be able to see printed on screen when I file is renamed.
- As a user of the libebpf library I want to be able to receive file rename events when calling `ebpf_event_ctx__next`

## Data needed

<table>
<tr>
	<td>Field
	<td>Type
	<td>Description
<tr>
	<td>File Name
	<td>char[256]
	<td>New name of the file being renamed 
<tr>
	<td>Full File Path (source)
	<td>char[4096]
	<td>Full source path of the file being renamed
<tr>
	<td>File File Path (destination)
	<td>char[4096]
	<td>Full destination path of the file being renamed
</table>

## Probe

As a starting point , some possible hooking points for this event could be:

-  fexit on `int do_renameat2(int olddfd, struct filename *from, int newdfd,
		 struct filename *to, unsigned int flags)`
- ~fexit on `int security_path_rename(const struct path *old_dir, struct dentry *old_dentry,
			 const struct path *new_dir, struct dentry *new_dentry,
			 unsigned int flags);` - needs a kernel compiled with `CONFIG_SECURITY_PATH`~ We can't do this, see this comment https://github.com/elastic/ebpf/issues/46#issuecomment-1000455585
- fexit on `int vfs_rename(struct renamedata *rd)`

## Action items

- [ ] Write one or more probes that can catch when a file is renamed
- [ ] Write BPF_PROG_TEST_RUN tests for the probes (if supported)
- [ ] Write integration tests for the probes
- [ ] Hook the event to be printed by the `EventsTrace` program

PR #67, opened 2022-01-25
EventProbe: implement EBPF_EVENT_FILE_RENAME

mmat11:
Closes https://github.com/elastic/ebpf/issues/44

### DIFF
--- a/GPL/EventProbe/EventProbe.bpf.c
+++ b/GPL/EventProbe/EventProbe.bpf.c
@@ -358,8 +358,8 @@ int BPF_PROG(fexit__vfs_rename)
     event->hdr.type = EBPF_EVENT_FILE_RENAME;
     event->hdr.ts   = bpf_ktime_get_ns();
     ebpf_pid_info__fill(&event->pids, task);
-    bpf_probe_read_str(event->old_path, PATH_MAX_BUF, s_state->rename.old_path);
-    bpf_probe_read_str(event->new_path, PATH_MAX_BUF, s_state->rename.new_path);
+    bpf_probe_read_kernel_str(event->old_path, PATH_MAX_BUF, s_state->rename.old_path);
+    bpf_probe_read_kernel_str(event->new_path, PATH_MAX_BUF, s_state->rename.new_path);
 
     bpf_ringbuf_submit(event, 0);
 
--- a/GPL/EventProbe/EventProbe.bpf.c
+++ b/GPL/EventProbe/EventProbe.bpf.c
@@ -284,11 +284,11 @@ int BPF_PROG(fentry__do_renameat2)
     ebpf_fileevents_state__set(EBPF_FILEEVENTS_STATE_RENAME, &state);
 
     u32 zero = 0;
-    struct ebpf_fileevents_scratch_state *s_state;
-    s_state = bpf_map_lookup_elem(&elastic_ebpf_fileevents_scratch_space, &zero);
-    if (!s_state)
+    struct ebpf_fileevents_scratch_space *ss =
+        bpf_map_lookup_elem(&elastic_ebpf_fileevents_init_buffer, &zero);
+    if (!ss)
         goto out;
-    ebpf_fileevents_scratch_state__set(EBPF_FILEEVENTS_STATE_RENAME, s_state);
+    ebpf_fileevents_scratch_space__set(EBPF_FILEEVENTS_STATE_RENAME, ss);
 
 out:
     return 0;
@@ -316,10 +316,10 @@ int BPF_PROG(fentry__vfs_rename)
         new_dentry = FUNC_ARG_READ(___type(new_dentry), vfs_rename, new_dentry);
     }
 
-    struct ebpf_fileevents_scratch_state *s_state =
-        ebpf_fileevents_scratch_state__get(EBPF_FILEEVENTS_STATE_RENAME);
-    if (!s_state) {
-        bpf_printk("fentry__vfs_rename: scratch state missing\n");
+    struct ebpf_fileevents_scratch_space *ss =
+        ebpf_fileevents_scratch_space__get(EBPF_FILEEVENTS_STATE_RENAME);
+    if (!ss) {
+        bpf_printk("fentry__vfs_rename: scratch space missing\n");
         goto out;
     }
 
@@ -328,9 +328,9 @@ int BPF_PROG(fentry__vfs_rename)
     struct path p;
     p.mnt    = state->rename.mnt;
     p.dentry = old_dentry;
-    ebpf_resolve_path_to_string(s_state->rename.old_path, &p, task);
+    ebpf_resolve_path_to_string(ss->rename.old_path, &p, task);
     p.dentry = new_dentry;
-    ebpf_resolve_path_to_string(s_state->rename.new_path, &p, task);
+    ebpf_resolve_path_to_string(ss->rename.new_path, &p, task);
 
     state->rename.step = RENAME_STATE_PATHS_SET;
 
@@ -352,10 +352,10 @@ int BPF_PROG(fexit__vfs_rename)
         goto out;
     }
 
-    struct ebpf_fileevents_scratch_state *s_state =
-        ebpf_fileevents_scratch_state__get(EBPF_FILEEVENTS_STATE_RENAME);
-    if (!s_state) {
-        bpf_printk("fexit__vfs_rename: scratch state missing\n");
+    struct ebpf_fileevents_scratch_space *ss =
+        ebpf_fileevents_scratch_space__get(EBPF_FILEEVENTS_STATE_RENAME);
+    if (!ss) {
+        bpf_printk("fexit__vfs_rename: scratch space missing\n");
         goto out;
     }
 
@@ -368,8 +368,8 @@ int BPF_PROG(fexit__vfs_rename)
     event->hdr.type = EBPF_EVENT_FILE_RENAME;
     event->hdr.ts   = bpf_ktime_get_ns();
     ebpf_pid_info__fill(&event->pids, task);
-    bpf_probe_read_kernel_str(event->old_path, PATH_MAX_BUF, s_state->rename.old_path);
-    bpf_probe_read_kernel_str(event->new_path, PATH_MAX_BUF, s_state->rename.new_path);
+    bpf_probe_read_kernel_str(event->old_path, PATH_MAX_BUF, ss->rename.old_path);
+    bpf_probe_read_kernel_str(event->new_path, PATH_MAX_BUF, ss->rename.new_path);
 
     bpf_ringbuf_submit(event, 0);
 
--- a/GPL/EventProbe/EventProbe.bpf.c
+++ b/GPL/EventProbe/EventProbe.bpf.c
@@ -79,7 +79,7 @@ int BPF_PROG(fexit__vfs_unlink)
     struct dentry *de        = NULL;
     struct task_struct *task = (struct task_struct *)bpf_get_current_task();
 
-    struct ebpf_fileevents_state *state;
+    struct ebpf_fileevents_state *state = NULL;
     state = ebpf_fileevents_state__get(EBPF_FILEEVENTS_STATE_UNLINK);
     if (state == NULL) {
         bpf_printk("fexit__vfs_unlink: no state\n");
--- a/GPL/EventProbe/EventProbe.bpf.c
+++ b/GPL/EventProbe/EventProbe.bpf.c
@@ -30,29 +30,41 @@
 
 char LICENSE[] SEC("license") = "GPL";
 
+/* vfs_unlink */
 DECL_RELO_FUNC_ARGUMENT(vfs_unlink, dentry);
 DECL_RELO_FUNC_RET(vfs_unlink);
+/* vfs_rename */
+DECL_RELO_FUNC_ARGUMENT(vfs_rename, old_dentry);
+DECL_RELO_FUNC_ARGUMENT(vfs_rename, new_dentry);
+DECL_RELO_FUNC_RET(vfs_rename);
 
 SEC("fentry/do_unlinkat")
 int BPF_PROG(fentry__do_unlinkat)
 {
-    struct ebpf_fileevents_tid_state state;
-    __builtin_memset(&state, 0, sizeof(struct ebpf_fileevents_tid_state));
-    state.state_id = EBPF_FILEEVENTS_TID_STATE_UNLINK;
-    ebpf_fileevents_write_state__set(&state);
+    struct ebpf_fileevents_state state;
+    __builtin_memset(&state, 0, sizeof(struct ebpf_fileevents_state));
+    ebpf_fileevents_state__set(EBPF_FILEEVENTS_STATE_UNLINK, &state);
     return 0;
 }
 
 SEC("fentry/mnt_want_write")
 int BPF_PROG(fentry__mnt_want_write, struct vfsmount *mnt)
 {
-    struct ebpf_fileevents_tid_state *state = ebpf_fileevents_write_state__get();
-    if (state == NULL)
-        goto out;
+    struct ebpf_fileevents_state *state;
 
-    struct ebpf_fileevents_unlink_state unlink_state;
-    unlink_state.mnt    = mnt;
-    state->state.unlink = unlink_state;
+    state = ebpf_fileevents_state__get(EBPF_FILEEVENTS_STATE_UNLINK);
+    if (state) {
+        state->unlink.mnt = mnt;
+        goto out;
+    }
+
+    state = ebpf_fileevents_state__get(EBPF_FILEEVENTS_STATE_RENAME);
+    if (state) {
+        state->rename.mnt  = mnt;
+        state->rename.step = RENAME_STATE_MOUNT_SET;
+        goto out;
+    }
+
 out:
     return 0;
 }
@@ -67,7 +79,8 @@ int BPF_PROG(fexit__vfs_unlink)
     struct dentry *de        = NULL;
     struct task_struct *task = (struct task_struct *)bpf_get_current_task();
 
-    struct ebpf_fileevents_tid_state *state = ebpf_fileevents_write_state__get();
+    struct ebpf_fileevents_state *state;
+    state = ebpf_fileevents_state__get(EBPF_FILEEVENTS_STATE_UNLINK);
     if (state == NULL) {
         bpf_printk("fexit__vfs_unlink: no state\n");
         goto out;
@@ -85,10 +98,9 @@ int BPF_PROG(fexit__vfs_unlink)
     event->hdr.ts   = bpf_ktime_get_ns();
     ebpf_pid_info__fill(&event->pids, task);
 
-    struct vfsmount *mnt = state->state.unlink.mnt;
     struct path p;
     p.dentry = de;
-    p.mnt    = mnt;
+    p.mnt    = state->unlink.mnt;
     ebpf_resolve_path_to_string(event->path, &p, task);
     bpf_ringbuf_submit(event, 0);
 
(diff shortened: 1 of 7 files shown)


GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -232,3 +244,93 @@ int BPF_PROG(fexit__do_filp_open,
...
+int BPF_PROG(fentry__do_renameat2)
+{
+    struct ebpf_fileevents_state state = {};
+    state.rename.step                  = RENAME_STATE_INIT;
+    ebpf_fileevents_state__set(EBPF_FILEEVENTS_STATE_RENAME, &state);
+    return 0;
+}
+
+SEC("fentry/vfs_rename")
+int BPF_PROG(fentry__vfs_rename)
```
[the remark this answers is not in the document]
nicholasberlin: It seems we could reduce the number of probes, remove the scratch space map, and only resolve paths when a rename was successful ... perhaps the old_path won't work anymore after `vfs_rename` completes?

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -232,3 +244,93 @@ int BPF_PROG(fexit__do_filp_open,
...
+{
+    struct ebpf_fileevents_state *state;
+    state = ebpf_fileevents_state__get(EBPF_FILEEVENTS_STATE_RENAME);
+    if (!state || state->rename.step != RENAME_STATE_MOUNT_SET) {
+        bpf_printk("fentry__vfs_rename: state missing or incomplete\n");
```
nicholasberlin, line 263: @mmat11 @fntlnz @leodido should we be removing entries from the maps? Both in error and success cases.

mmat11: IIRC the `LRU_HASH` map should clean up itself

nicholasberlin: So we just let it grow to max size, and let it pops things as needed?

GPL/EventProbe/EventProbe.bpf.c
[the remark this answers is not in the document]
nicholasberlin: though, I wonder how the path traversal works for the new path if it doesn't yet exist.

nicholasberlin (approved):
LGTM

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -30,29 +30,41 @@
...
+    ebpf_fileevents_state__set(EBPF_FILEEVENTS_STATE_UNLINK, &state);
     return 0;
 }
 
 SEC("fentry/mnt_want_write")
 int BPF_PROG(fentry__mnt_want_write, struct vfsmount *mnt)
 {
-    struct ebpf_fileevents_tid_state *state = ebpf_fileevents_write_state__get();
-    if (state == NULL)
+    struct ebpf_fileevents_state *state;
```
fntlnz, line 53:
```suggestion
    struct ebpf_fileevents_state *state;
```
```suggestion
    struct ebpf_fileevents_state *state = NULL;
```

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -67,7 +79,8 @@ int BPF_PROG(fexit__vfs_unlink)
     struct dentry *de        = NULL;
     struct task_struct *task = (struct task_struct *)bpf_get_current_task();
 
-    struct ebpf_fileevents_tid_state *state = ebpf_fileevents_write_state__get();
+    struct ebpf_fileevents_state *state;
```
fntlnz, line 82:
```suggestion
    struct ebpf_fileevents_state *state;
```
```suggestion
    struct ebpf_fileevents_state *state = NULL;
```

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -232,3 +244,93 @@ int BPF_PROG(fexit__do_filp_open,
...
+{
+    struct ebpf_fileevents_state *state;
+    state = ebpf_fileevents_state__get(EBPF_FILEEVENTS_STATE_RENAME);
+    if (!state || state->rename.step != RENAME_STATE_MOUNT_SET) {
+        bpf_printk("fentry__vfs_rename: state missing or incomplete\n");
+        goto out;
+    }
+
+    struct dentry *old_dentry, *new_dentry;
+    if (bpf_core_type_exists(struct renamedata)) {
```
fntlnz, line 268: The builtin `__builtin_preserve_type_info` used by this does not work with the current toolchain.

We have two options:

- update the toolchain
- add a new macro like `RELO_FENTRY_ARG_EXISTS` that checks if a function argument is present.


We should probably do both regardless, to make this implementation stronger

fntlnz (commented):
Thanks for working on this @mmat11 

Did an initial pass and had some comments.

GPL/EventProbe/EventProbe.bpf.c
mmat11: > add a new macro like RELO_FENTRY_ARG_EXISTS that checks if a function argument is present

Sounds good to me, what about naming? these macros are used only in fentry/fexit prog types for now but the userspace code is not directly related to the prog type so we could omit that.
For this specific one, `RELO` doesn't sound convincing aswell :D; what about `FUNC_HAS_ARG`/`FUNC_PROTO_HAS_ARG` or something like that?

fntlnz: I like the names you proposed better as well.

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -232,3 +244,93 @@ int BPF_PROG(fexit__do_filp_open,
...
+}
+
+SEC("fexit/vfs_rename")
+int BPF_PROG(fexit__vfs_rename)
+{
+    int ret = RELO_FENTRY_RET_READ(___type(ret), vfs_rename);
+    if (ret)
+        goto out;
+
+    struct ebpf_fileevents_state *state;
```
fntlnz, line 307:
```suggestion
    struct ebpf_fileevents_state *state = NULL;
```

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -73,102 +73,135 @@ const struct btf_type *resolve_btf_type_by_func_name(struct btf *btf, const char
...
         __r;                                                                                       \
     })
 
+/* Given a function name and an argument name, returns whether the argument
+ * exists or not.
+ */
+#define FILL_FUNC_ARG_EXISTS(ctx, btf, func, arg)                                                  \
+    ({                                                                                             \
+        bool __r = false;                                                                          \
+        int _r   = resolve_btf_func_arg_exists(btf, #func, #arg);                                  \
```
fntlnz, line 162: I've been thinking we could just use `resolve_btf_func_arg_idx` and check if the return value is `< 0` (does not exist) or `>= 0` exists. Wdyt?

mmat11: good idea!

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -72,89 +73,112 @@ const struct btf_type *resolve_btf_type_by_func_name(struct btf *btf, const char
...
+
+    /* The following two macros can fail as the arguments may not exist in recent signatures.
+     * The indexes will be used only in the case `struct renamedata` is not present.
+     */
+    FILL_FUNC_ARG_IDX(ctx, btf, vfs_rename, old_dentry);
+    FILL_FUNC_ARG_IDX(ctx, btf, vfs_rename, new_dentry);
+    err = FILL_FUNC_RET_IDX(ctx, btf, vfs_rename);
     if (err)
         goto out;
+    err = FILL_FUNC_ARG_EXISTS(ctx, btf, vfs_rename, rd);
```
nicholasberlin, line 176: This macro checks `>= 0` should line 177 check for less than zero?

mmat11: yes, actually I had the same bug before and I thought I got rid of it :D; I'll change the macro to be consistent with the other macros in terms of errors

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -72,89 +73,112 @@ const struct btf_type *resolve_btf_type_by_func_name(struct btf *btf, const char
...
 
+/* Given a function name and an argument name, returns whether the argument
+ * exists or not.
+ */
+#define FILL_FUNC_ARG_EXISTS(ctx, btf, func, arg)                                                  \
+    ({                                                                                             \
+        bool __r = false;                                                                          \
+        int _r   = resolve_btf_func_arg_idx(btf, #func, #arg);                                     \
+        if (_r >= 0)                                                                               \
+            __r = true;                                                                            \
```
nicholasberlin, line 141: I think we should set `_r = 0` here. The use of the macros seems to assume anything other than 0 being returned is an error.

mmat11: Should be fixed in the latest commit

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -135,12 +136,13 @@ static int resolve_btf_func_ret_idx(struct btf *btf, const char *func)
...
-        bool __r = false;                                                                          \
-        int _r   = resolve_btf_func_arg_idx(btf, #func, #arg);                                     \
-        if (_r >= 0)                                                                               \
-            __r = true;                                                                            \
-        (*ctx)->probe->rodata->exists__##func##__##arg##__ = __r;                                  \
-        _r;                                                                                        \
+        int __r = -1;                                                                              \
+        int r   = resolve_btf_func_arg_idx(btf, #func, #arg);                                      \
+        if (r >= 0) {                                                                              \
+            (*ctx)->probe->rodata->exists__##func##__##arg##__ = true;                             \
```
nicholasberlin, line 142: This was set unconditionally before, does it default to false?

mmat11: It gets declared here https://github.com/elastic/ebpf/blob/973292635388fd2a940b7c15b62480d09398fc6e/GPL/EventProbe/Helpers.h#L47

nicholasberlin: Ah, cool. Thanks.

nicholasberlin (approved):
I already approved this, but the recent changes LGTM.

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -232,3 +245,93 @@ int BPF_PROG(fexit__do_filp_open,
...
+    if (!event)
+        goto out;
+
+    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+
+    event->hdr.type = EBPF_EVENT_FILE_RENAME;
+    event->hdr.ts   = bpf_ktime_get_ns();
+    ebpf_pid_info__fill(&event->pids, task);
+    bpf_probe_read_str(event->old_path, PATH_MAX_BUF, s_state->rename.old_path);
+    bpf_probe_read_str(event->new_path, PATH_MAX_BUF, s_state->rename.new_path);
```
rhysre, line 331: Should be `bpf_probe_read_kernel_str` to be in-line with current kernel recommendations.

From `man bpf_helpers`:


```
long bpf_probe_read_str(void *dst, u32 size, const void
       *unsafe_ptr)

              Description
                     Copy a NUL terminated string from an unsafe kernel
                     address unsafe_ptr to dst. See
                     bpf_probe_read_kernel_str() for more details.

                     Generally, use bpf_probe_read_user_str() or
                     bpf_probe_read_kernel_str() instead.

              Return On success, the strictly positive length of the
                     string, including the trailing NUL character. On
                     error, a negative value.
```

GPL/EventProbe/Helpers.h
```diff
@@ -28,23 +28,25 @@
...
+#define FUNC_RET_READ(type, func)                                                                  \
     ({                                                                                             \
         type _ret;                                                                                 \
-        bpf_core_read(&_ret, sizeof(_ret), ctx + ret__##func_name##__);                            \
+        bpf_core_read(&_ret, sizeof(_ret), ctx + ret__##func##__);                                 \
         _ret;                                                                                      \
     })
 
+#define DECL_FUNC_ARG_EXISTS(func, arg) const volatile bool exists__##func##__##arg##__ = false;
+#define FUNC_ARG_EXISTS(func, arg) exists__##func##__##arg##__
```
rhysre, line 48: Really liking how bulletproof and extensible BTF is making things here. Good work @fntlnz and @mmat11 🚀

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -232,3 +245,93 @@ int BPF_PROG(fexit__do_filp_open,
...
+    if (!state || state->rename.step != RENAME_STATE_MOUNT_SET) {
+        bpf_printk("fentry__vfs_rename: state missing or incomplete\n");
+        goto out;
+    }
+
+    struct dentry *old_dentry, *new_dentry;
+    if (FUNC_ARG_EXISTS(vfs_rename, rd)) {
+        /* Function arguments have been refactored into struct renamedata */
+        struct renamedata *rd = (struct renamedata *)ctx[0];
+        old_dentry            = rd->old_dentry;
```
rhysre, line 272: The old kernel headers in the toolchain are becoming a nuisance again, this can't be built with the toolchain:

```
/home/vagrant/endpoint-dev/Kernel/Ebpf/GPL/EventProbe/EventProbe.bpf.c:272:35: error: incomplete definition of type 'struct renamedata'
        old_dentry            = rd->old_dentry;
                                ~~^
/home/vagrant/endpoint-dev/Kernel/Ebpf/GPL/EventProbe/EventProbe.bpf.c:271:16: note: forward declaration of 'struct renamedata'
        struct renamedata *rd = (struct renamedata *)ctx[0];
               ^
/home/vagrant/endpoint-dev/Kernel/Ebpf/GPL/EventProbe/EventProbe.bpf.c:273:35: error: incomplete definition of type 'struct renamedata'
        new_dentry            = rd->new_dentry;
                                ~~^
/home/vagrant/endpoint-dev/Kernel/Ebpf/GPL/EventProbe/EventProbe.bpf.c:271:16: note: forward declaration of 'struct renamedata'
        struct renamedata *rd = (struct renamedata *)ctx[0];
               ^
```

`struct renamedata` was introduced in Linux 5.12 (torvalds/linux@9fe61450972d3900bffb1dc26a17ebb9cdd92db2) and so the 3.x headers in the toolchain don't contain it. This is going to break the build with endpoint as-is. We might have to hack our way around this by redefining the struct.

I assume you're not using the toolchain to build this and are on a 5.12+ box @mmat11 ?

GPL/EventProbe/FileEventsHelpers.h
```diff
@@ -21,39 +21,97 @@
...
+};
+
+/* Use a scratch map to store big objects as a state in
+ * prog chains.
+ *
+ * https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net.git/commit/?id=5722569bb9c3bd922c4f10b5b2912fe88c255312
+ */
+struct bpf_map_def SEC("maps") elastic_ebpf_fileevents_scratch_state = {
+    .type        = BPF_MAP_TYPE_PERCPU_ARRAY,
+    .key_size    = sizeof(enum ebpf_fileevents_scratch_key),
```
rhysre, line 112: Why does this have to be indexed by `ebpf_fileevents_scratch_key`? There should only ever be one file operation at a time ongoing on a single CPU core, right? So why is the indexing needed? Why not just look it up in this map with a constant key `0` regardless of the operation every time?

rhysre (commented):
A few comments. Looking really good in general, nice work 🚀 

I'm concerned that this breaks the build with the current toolchain, that's going to totally block us when we want to integrate it with endpoint. I think we'll unfortunately need to copy more definitions in from kernel headers until we've dockerized the build process and are using modern kernel headers.

GPL/EventProbe/EventProbe.bpf.c
mmat11: nice catch, I always get confused

GPL/EventProbe/FileEventsHelpers.h
mmat11: Correct me if I am wrong, to my understanding, the area between 2 ebpf hooks excluding the ebpf program execution itself, can be preemptible, so it could happen that the same CPU could execute another ebpf program while the state is not read yet

> There should only ever be one file operation at a time ongoing on a single CPU core, right?

I don't know about file operations in particular, but I didn't want to complicate things and I thought that having one key per operation would make things less confusing at the cost of maybe more `max_entries` in the future. My assumption is that we shouldn't need many entries in this map so it should be fine

GPL/EventProbe/EventProbe.bpf.c
mmat11: > I assume you're not using the toolchain to build this and are on a 5.12+ box @mmat11 ?

Correct, @fntlnz told me this yesterday and my first task today was to recompile with the toolchain and fix anything that comes out

mmat11: I was able to build with the toolchain, I also triggered a CI run and couldn't see these specific errors

rhysre: I was mistaken in my original comment when I referenced the kernel headers in the toolchain.

It's not the toolchain headers. I got this error because I was building on vm with a 5.11 kernel, and thus the generated `vmlinux.h` didn't have the 5.12+ `struct renamedata` in it. I'm assuming your local machine is 5.12+?

The "builtin" vmlinux used to build stuff in jenkins is from 5.12 or newer, and thus has that struct. Workaround for those on pre 5.11 is to just use `-DUSE_BUILTIN_VMLINUX` instead.

Worth noting that, with the current endpoint build setup (which generates a vmlinux.h based on the machine's kernel), endpoint will fail to build on Linux if the buildbox is pre-5.12 (unless we switch to the builtin vmlinux). But that's a solveable problem (upgrade your vm) instead of a hard problem baked into the toolchain. I think this is fine for now, worst case scenario in the future -- everybody uses a prebuilt vmlinux.h.

@fntlnz Are there any issues with using a prebuilt `vmlinux.h` by default for the build to avoid this problem entirely? I foresee a bunch of angry devs if their endpoint builds suddently break if they're using a pre-5.12 kernel.

GPL/EventProbe/FileEventsHelpers.h
rhysre: > Correct me if I am wrong, to my understanding, the area between 2 ebpf hooks excluding the ebpf program execution itself, can be preemptible

I don't _think_ code is _preemptable_ if it's running in kernelspace (AFAIK that would break a number of kernel locking assumptions), but the possibility exists for it to voluntarily sleep and be put in a waitqueue, which is for our purposes the same thing as it would lead to something else being scheduled on that core. My knowledge here isn't super strong, maybe someone else can chime in (@fntlnz ?)

Unsure if that can happen here with `vfs_rename`, but assuming it does, the problem you're trying to solve is valid (I didn't consider sleeping in my original comment).

This code still would have a problem however, consider the following sequence of events with the current code:
1. fentry on `vfs_rename` fires on core 1, state is put in the `core#1` slot of the `scratch_state` map
2. `vfs_rename` resumes
3. `vfs_rename` sleeps
4. `vfs_rename` wakes up and is scheduled on core 2
5. `fexit` on `vfs_rename` fires, state is read from the `core#2` slot of the `scratch_state` map (wrong)

Using a percpu map would break if any file operation was rescheduled to a new core in the mid-kernel-execution. What we should be doing is making this an LRU map and indexing it by `ebpf_fileevents_state__key`, just like  the `fileevents_state` map. That way, regardless of what core, or what sequence of cores a file operation runs on, it will always read the same value from the scratch map as it will be indexed by tid/tgid and not cpu core.

rhysre (commented):
Few more comments and follow ups

GPL/EventProbe/EventProbe.bpf.c
mmat11: > I think this is fine for now, worst case scenario in the future -- everybody uses a prebuilt vmlinux.h.

what about keeping an updated vmlinux.h and only redefine structs that have been removed in newer versions once we go for the update?

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -72,89 +73,114 @@ const struct btf_type *resolve_btf_type_by_func_name(struct btf *btf, const char
...
+    err = FILL_FUNC_ARG_IDX(ctx, btf, vfs_unlink, dentry);
     if (err)
         goto out;
-    err = FILL_FUNCTION_RET_RELO(ctx, vfs_unlink);
+
+    err = FILL_FUNC_RET_IDX(ctx, btf, vfs_unlink);
+    if (err)
+        goto out;
+
+    /* The following two macros can fail as the arguments may not exist in recent signatures.
```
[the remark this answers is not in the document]
mmat11: nice catch 👍🏻 , I'll move the "exists" macro earlier so we can check errors on the other macros

GPL/EventProbe/EventProbe.bpf.c
fntlnz: @mmat11 @rhysre When I converted this project to cmake originally did `-DUSE_BUILTIN_VMLINUX=False` because it was the behavior pre-cmake. We can probably just generate it once but we would need to agree on a kernel to be used as base, document that and when we need to update we need to agree on the newer kernel.

Otheriwise we generate it from our laptops based on the distro we personally use. I'd rather have a standard thing for all of us.

GPL/EventProbe/FileEventsHelpers.h
fntlnz: Good intuition @rhysre - I had the same doubt when doing the support map for file unlink but never took a chance to go deeper.

I've been trying to elaborate. Here are some thoughts

The linux kernel only operates in two contexts, behalf of a process or in atomic context.

When executing a syscall, the only operation under a spinlock is the IRQ handling and the rest of the code is sleepable

Interrupt signal trigger 
image

Interrupt handler (code at https://elixir.bootlin.com/linux/latest/source/kernel/irq/chip.c#L780)

image


So yeah, there are no guarantees that this code does not sleep.

GPL/EventProbe/EventProbe.bpf.c
mmat11: I've opened https://github.com/elastic/ebpf/issues/69 to keep track of this, resolving the discussion for now

GPL/EventProbe/FileEventsHelpers.h
mmat11: I've tried to use a normal map but I couldn't use the `ebpf_fileevents_state__key` as I need to init the value and take the space for it in the ebpf program. The alternatives are: use an array and keep track of the used indexes in another map and clean it up once the chain is finished (but this also has the problem of keeping the two maps consistent with each other) or accept the possibility that a task could be cpu-migrated between two programs (I've had a look at some percpu-refcounters https://elixir.bootlin.com/linux/v5.16.4/source/kernel/bpf/trampoline.c#L246 in trampoline.c but I am not sure about how is that being used in this regard; I can't confirm that a cpu migration can't happen)

mmat11: I solved by using the percpu-map only for the init of the lru hash; let me know if the scratch map size is ok for you

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -249,9 +249,17 @@ int BPF_PROG(fexit__do_filp_open,
 SEC("fentry/do_renameat2")
 int BPF_PROG(fentry__do_renameat2)
 {
-    struct ebpf_fileevents_state state = {};
-    state.rename.step                  = RENAME_STATE_INIT;
+    struct ebpf_fileevents_state state = {.rename = {.step = RENAME_STATE_INIT}};
     ebpf_fileevents_state__set(EBPF_FILEEVENTS_STATE_RENAME, &state);
```
rhysre, line 253: I'm having a very hard time understanding this code, perhaps there's something I'm missing. Why do we need the percpu map? Why not just call `ebpf_fileevents_state__set` directly and ditch the percpu map entirely?

mmat11: We are using 3 maps here:

- `elastic_ebpf_fileevents_state` (LRU_HASH)
- `elastic_ebpf_fileevents_scratch_space` (PERCPU_ARRAY)
- `elastic_ebpf_fileevents_scratch_state` (LRU_HASH)

The percpu map is being used in the context of a single program to initialize `ebpf_fileevents_scratch_state` as doing that in a non-array map or the program itself always raises a verifier error related to the stack size.

I also made the third map (scratch_state) so we can set a lower amount of `max_entries` as we don't need this space for all the programs that uses the other LRU_HASH

rhysre: Ahhhh yes, ok now I understand. We had to solve the same issue with the Cmd probes, this is all coming back. In the Cmd probes we just called this the "rename buffer" because it was only used in renames.

I think my confusion stemmed from the naming of these three maps. IMO the current naming is pretty confusing. Can we rename them as follows:

- `elastic_ebpf_fileevents_scratch_space` -> `initialization_buffer_map` (and put a comment above it explaining its why it's necessary due to the 512 byte BPF stack)
- `elastic_ebpf_fileevents_scratch_state` -> `elastic_ebpf_fileevents_scratch_space` (also update the struct definitions `s/state/space/g`. having both "scratch state" and "scratch space" is really confusing IMO, "scratch space" is the much more widely used term)

mmat11: > the current naming is pretty confusing

agree 👍🏻, will change the names tomorrow

fntlnz (approved):

Issue #68, opened 2022-01-25
Use BTF-defined maps


mmat11:
Legacy definitions will be deprecated soon and probably not supported anymore by 1.0.

https://lore.kernel.org/bpf/20220120060529.1890907-1-andrii@kernel.org/

leodido:
+1 as discussed privately

PR #71, opened 2022-02-02
Use BTF-defined map defs and change PathResolver scratch map to a percpu-array

mmat11:
Closes https://github.com/elastic/ebpf/issues/68

@rhysre I've changed the scratch map in PathResolver to a percpu-array since the execution is always in the context of a single ebpf program. Let me know if that's ok for you


rhysre (approved):
Nice catch on the pathresolver scratch map not needing to be an array 👍

fntlnz (approved):
LGTM

PR #72, opened 2022-02-03
EventProbe: split EventProbe.bfc.c file into File, Process and Network

mmat11:
This keeps a single `.o` bpf object for now, it's just for readability


GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -23,356 +23,19 @@
...
-}
+// todo(fntlnz): another buffer will probably need
+// to be used instead of this one as the common parts evolve
+// to have a shared buffer between File, Network and Process.
+struct {
+    __uint(type, BPF_MAP_TYPE_RINGBUF);
+    __uint(max_entries, 4096 * 64); // todo: Need to verify if 256 kb is what we want
+} ringbuf SEC(".maps");
+
+#include "File/Probe.h"
```
nicholasberlin, line 39: Perhaps these should be renamed to, for example, `File/Probe.bpf.c`. Because there aren't really headers, right?

Looks like CMakeLists.txt only builds `GPL/EventProbe/EventProbe.bpf.c`, so new `*.bpf.c` files won't be picked up by a file glob.

GPL/EventProbe/EventProbe.bpf.c
```diff
@@ -23,356 +23,19 @@
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_tracing.h>
```
rhysre, line 24: These three includes don't need to be in the top level `.c` file anymore, can they be moved to the specific files under Process and File that they're needed in.

GPL/EventProbe/EventProbe.bpf.c
rhysre: The distinction between headers and source files becomes a bit muddied in BPF-land. This is essentially a stylistic choice, but IMO I agree that calling the files with the actual probe definitions `.bpf.c` and then including them all in `EventProbe.bpf.c` is cleaner. @mmat11 what are your thoughts?

mmat11: > @mmat11 what are your thoughts?

I initially called them `Probe.c`, then I changed a couple times (`.bpf.c`, `.h`) and didn't like any, `.bpf.c` sounds better than `.h`, so I'll change

GPL/EventProbe/EventProbe.bpf.c
mmat11: what about `vmlinux.h`? I think that's the only one that is going to get included in every "Probe.bpf.c", right?

GPL/EventProbe/EventProbe.bpf.c
mmat11: Changed to `.bpf.c`

GPL/EventProbe/EventProbe.bpf.c
rhysre: True. I would however add three separate `#include "vmlinux.h"` directives to each Probe.bpf.c file instead of just including it in `EventProbe.bpf.c`. 

Those probes should (in theory) be able to be built on their own, having them all rely on the transitive include strikes me as unclean.

I think you're also going to have to include it in `EventProbe.bpf.c` due to it requiring the `BPF_MAP_TYPE_RINGBUF` definition from vmlinux.

so tl;dr IMO include `vmlinux.h` here in `EventProbe.bpf.c` and all probe files in each subdirectory.

mmat11: done

rhysre (approved):

nicholasberlin (approved):

Issue #63, opened 2022-01-18
Network Events: Implement  EBPF_EVENT_NETWORK_IPV4_DISCONNECT_RECEIVED


fntlnz:
## Stories

- As a user of the `EventsTrace` program I want to be able to see printed on screen when an IPv4 Disconnect is received
- As a user of the libebpf library I want to be able to receive ipv4 disconnection received events when calling `ebpf_event__next`

# Data Needed

TODO: write this

Issue #62, opened 2022-01-18
Network Events: Implement EBPF_EVENT_NETWORK_IPV4_CONNECTION_ACCEPTED


fntlnz:
## Stories

- As a user of the `EventsTrace` program I want to be able to see printed on screen when an IPv4 Connection is accepted
- As a user of the libebpf library I want to be able to receive ipv4 connection accepted events when calling `ebpf_event__next`

# Data Needed

TODO: write this

Issue #61, opened 2022-01-18
Network Events: Implement EBPF_EVENT_NETWORK_IPV4_CONNECTION_ATTEMPTED


fntlnz:
## Stories

- As a user of the `EventsTrace` program I want to be able to see printed on screen when an IPv4 Connection attempt happens
- As a user of the libebpf library I want to be able to receive ipv4 connection attempt events when calling `ebpf_event_ctx__next`

## Data needed

<table>
<tr>
	<td>Field
	<td>Type
	<td>Description
<tr>
	<td>Destination port
	<td>uint32_t
	<td>Destination port of the attempted connection 
<tr>
	<td>Destination Address
	<td>char[16]
	<td>String representation of the IPv4 of the destination
<tr>
	<td>Socket pointer (sk)
	<td>uint64_t
	<td>The kernel socket pointer
</table>

## Probe
Some possible approaches could be:

- fexit on `int inet_stream_connect(struct socket *sock, struct sockaddr *uaddr,
			int addr_len, int flags)`

## Action items

- [ ] Write one or more probes that can catch when an IPv4 connection is attempted
- [ ] Write BPF_PROG_TEST_RUN tests for the probes (if supported)
- [ ] Write integration tests for the probes
- [ ] Hook the event to be printed by the `EventsTrace` program

PR #74, opened 2022-02-06
Network events

mmat11:
To do / discuss:
- [x] clarify what "disconnect received" means: [tcp_close](https://elixir.bootlin.com/linux/v5.16.7/source/net/ipv4/tcp_ipv4.c#L3058) or [tcp_disconnect](https://elixir.bootlin.com/linux/v5.16.7/source/net/ipv4/tcp_ipv4.c#L3061) ? rename event?
- [x] clarify what "connection attempted" means: only successful attempts? rename event?
- [x] test ipv6
- [x] check if any func signature has changed from 5.x to 5.16
- [x] doing stuff in a container generates 2 events (host netns and container netns), eg:
```
{"family":"AF_INET","source_address":"10.0.2.100","source_port":42270,"destination_address":"157.90.119.155","destination_port":443,"network_namespace":4026534259}
{"family":"AF_INET","source_address":"192.168.1.3","source_port":43526,"destination_address":"157.90.119.155","destination_port":443,"network_namespace":4026532008}
``` 
Why it happened: we debugged this with @fntlnz and discovered that the second event was being triggered by [slirp4netns](https://github.com/rootless-containers/slirp4netns) because I was using Podman in rootless mode. This doesn't happen using Docker

- [x] add `bytes_sent` and `bytes_received` on `tcp_close`
- [x] add transport field

Closes https://github.com/elastic/ebpf/issues/61
Closes https://github.com/elastic/ebpf/issues/62
Closes https://github.com/elastic/ebpf/issues/63

### DIFF
--- a/GPL/EventProbe/EbpfEventProto.h
+++ b/GPL/EventProbe/EbpfEventProto.h
@@ -33,6 +33,7 @@ enum ebpf_event_type {
     EBPF_EVENT_FILE_RENAME                  = (1 << 7),
     EBPF_EVENT_NETWORK_CONNECTION_ACCEPTED  = (1 << 8),
     EBPF_EVENT_NETWORK_CONNECTION_ATTEMPTED = (1 << 9),
+    EBPF_EVENT_NETWORK_CONNECTION_CLOSED    = (1 << 10),
 };
 
 struct ebpf_event_header {
--- a/GPL/EventProbe/EbpfEventProto.h
+++ b/GPL/EventProbe/EbpfEventProto.h
@@ -24,13 +24,14 @@
 #endif
 
 enum ebpf_event_type {
-    EBPF_EVENT_PROCESS_FORK   = (1 << 1),
-    EBPF_EVENT_PROCESS_EXEC   = (1 << 2),
-    EBPF_EVENT_PROCESS_EXIT   = (1 << 3),
-    EBPF_EVENT_PROCESS_SETSID = (1 << 4),
-    EBPF_EVENT_FILE_DELETE    = (1 << 5),
-    EBPF_EVENT_FILE_CREATE    = (1 << 6),
-    EBPF_EVENT_FILE_RENAME    = (1 << 7),
+    EBPF_EVENT_PROCESS_FORK                = (1 << 1),
+    EBPF_EVENT_PROCESS_EXEC                = (1 << 2),
+    EBPF_EVENT_PROCESS_EXIT                = (1 << 3),
+    EBPF_EVENT_PROCESS_SETSID              = (1 << 4),
+    EBPF_EVENT_FILE_DELETE                 = (1 << 5),
+    EBPF_EVENT_FILE_CREATE                 = (1 << 6),
+    EBPF_EVENT_FILE_RENAME                 = (1 << 7),
+    EBPF_EVENT_NETWORK_CONNECTION_ACCEPTED = (1 << 8),
 };
 
 struct ebpf_event_header {
@@ -106,6 +107,32 @@ struct ebpf_process_exit_event {
 struct ebpf_process_setsid_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
+} __attribute__((packed));
+
+enum ebpf_net_info_af {
+    EBPF_NETWORK_EVENT_AF_INET  = 1,
+    EBPF_NETWORK_EVENT_AF_INET6 = 2,
+};
+
+struct ebpf_net_info {
+    enum ebpf_net_info_af family;
+    union {
+        uint8_t saddr[4];
+        uint8_t saddr6[16];
+    }; // Network byte order
+    union {
+        uint8_t daddr[4];
+        uint8_t daddr6[16];
+    }; // Network byte order
+    uint16_t sport;
+    uint16_t dport;
+    uint32_t netns;
+} __attribute__((packed));
+
+struct ebpf_net_connection_accepted_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    struct ebpf_net_info net;
 } __attribute__((packed));
 
 #endif // EBPF_EVENTPROBE_EBPFEVENTPROTO_H
--- a/GPL/EventProbe/EbpfEventProto.h
+++ b/GPL/EventProbe/EbpfEventProto.h
@@ -111,6 +111,10 @@ struct ebpf_process_setsid_event {
     struct ebpf_pid_info pids;
 } __attribute__((packed));
 
+enum ebpf_net_info_transport {
+    EBPF_NETWORK_EVENT_TRANSPORT_TCP = 1,
+};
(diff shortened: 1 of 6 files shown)


nicholasberlin (commented):
I don't know if it will cause issues here or not, but in the past I had trouble with programs that attempted to connect with using a non blocking socket. The connect syscall would return with `-EINPROGRESS`, which would later succeed.

I'm pretty sure curl uses that style.

mmat11:
> I don't know if it will cause issues here or not, but in the past I had trouble with programs that attempted to connect using a non blocking socket. The connect syscall would return with `-EINPROGRESS`, which would later succeed.
> 
> I'm pretty sure curl uses that style.

I tried to reproduce this way:

```
import socket

sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.connect(('localhost', 1234))
sock.setblocking(0)  // O_NONBLOCK

sock.send(b'hello\n')
```
and
```
nc -lvp 1234
Ncat: Version 7.91 ( https://nmap.org/ncat )
Ncat: Listening on :::1234
Ncat: Listening on 0.0.0.0:1234
Ncat: Connection from 127.0.0.1.
Ncat: Connection from 127.0.0.1:56718.
hello
```
which triggered this event from v4_connect
```
{"event_type":"NETWORK_CONNECTION_ATTEMPTED","pids":{"tid":109704,"tgid":109704,"ppid":74299,"pgid":109704,"sid":74299,"start_time_ns":21219042657261},"net":{"family":"AF_INET","source_address":"127.0.0.1","source_port":56714,"destination_address":"127.0.0.1","destination_port":1234,"network_namespace":4026532008},"comm":"python"}
```

I also didn't see any event missing during other tests with curl, so I think it should be fine

GPL/EventProbe/Network/Probe.bpf.c
```diff
@@ -17,3 +17,96 @@
...
+int BPF_PROG(fexit__tcp_v6_connect, struct sock *sk, struct sockaddr *uaddr, int addr_len, int ret)
+{
+    return tcp_connect(sk, ret);
+}
+
+SEC("fentry/tcp_close")
+int BPF_PROG(fentry__tcp_close, struct sock *sk, long timeout)
+{
+    unsigned char state = BPF_CORE_READ(sk, __sk_common.skc_state);
+    if (state == TCP_CLOSE)
```
mmat11, line 91: this is not a good solution, between `bind()`/`connect()` and `tcp_close()`, other functions may set the sk state to `TCP_CLOSE` (eg. `tcp_reset()`->`tcp_done()`) and the event is lost

non-GPL/EventsTrace/EventsTrace.c
```diff
@@ -314,6 +327,115 @@ static void out_process_exit(struct ebpf_process_exit_event *evt)
...
+        break;
+    }
+
+    out_comma();
+    out_int("network_namespace", net->netns);
+
+    switch (evt->hdr.type) {
+    case EBPF_EVENT_NETWORK_CONNECTION_CLOSED:
+        out_comma();
+        out_int("bytes_sent", net->tcp.close.bytes_sent);
```
stanek-michal, line 396: out_uint I suppose

GPL/EventProbe/EbpfEventProto.h
```diff
@@ -108,4 +111,44 @@ struct ebpf_process_setsid_event {
...
+    union {
+        uint8_t daddr[4];
+        uint8_t daddr6[16];
+    }; // Network byte order
+    uint16_t sport;
+    uint16_t dport;
+    uint32_t netns;
+    union {
+        struct ebpf_net_info_tcp_close close;
+    } tcp;
```
stanek-michal, line 144: What else will we have in this union? Do we need the union at all? I'm not sure about the naming. Maybe tcp_stats or tcp_close_stats would be better?

GPL/EventProbe/Network/Probe.bpf.c
```diff
@@ -17,3 +17,96 @@
...
+    struct ebpf_net_event *event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
+    if (!event)
+        goto out;
+
+    if (ebpf_network_event__fill(event, sk)) {
+        bpf_ringbuf_discard(event, 0);
+        goto out;
+    }
+
+    event->hdr.type = EBPF_EVENT_NETWORK_CONNECTION_ATTEMPTED;
```
stanek-michal, line 68: I wonder if CONNECTION_ESTABLISHED wouldn't be a better name. Unless it's already set in stone somewhere.

stanek-michal: Oh right, tcp_v4_connect returning 0 doesn't mean the connection is established. Nevermind!

GPL/EventProbe/Network/Probe.bpf.c
stanek-michal: hm, I wonder why do we want to check this flag at all?

mmat11: I'd like to drop "empty" events generated by stuff like this:

```
import socket
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.close()
```

`nc` does this aswell

GPL/EventProbe/EbpfEventProto.h
mmat11: The union is for event specific info (in this case tcp), I'd like to keep fields like this separated from the main net info as they are specific to certain events. In the future, I suppose there will also be a `udp` union, which will probably be merged and renamed into this one

GPL/EventProbe/Network/Network.h
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+#ifndef EBPF_EVENTPROBE_NETWORK_H
+#define EBPF_EVENTPROBE_NETWORK_H
+
+// linux/socket.h
+#define AF_INET 2
+#define AF_INET6 10
+
+static int ebpf_sock_info__fill(struct ebpf_net_info *net, struct sock *sk)
```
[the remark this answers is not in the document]
mmat11: They can be moved but I think they are more organized this way. I don't have a strong opinion on this, so I'll move them unless any other reviewer also likes the separate header file

GPL/EventProbe/Network/Probe.bpf.c
mmat11: > Unless it's already set in stone somewhere

IMO `elastic/ebpf` has its own domain, so I don't think we have to match 1:1 existing event names - I suppose there will be a "translation" layer

stanek-michal (approved):
LGTM, thanks for the changes!

rhysre:
Was OOO last week so didn't see this, but just looked it over and wanted to say good job @mmat11! This is really useful data from a security standpoint and I'm excited to see us generating it in BPF!

PR #76, opened 2022-02-10
Default to using builtin vmlinux

mjwolf:
Set USE_BUILTIN_VMLINUX to default to True.

On systems with kernels that don't provide BTF info, the build will fail
when the builtin vmlinux is not used. So defaulting to use builtin
vmlinux is the safer/more compatible build option.


rhysre (approved):
We're going to need to use the builtin vmlinux.h when building this repo with endpoint (otherwise building on Linux will result in errors on pre-5.12). 

I suppose we could set this option explicitly in the endpoint repo, but I'm partial to just using the builtin vmlinux.h by default as Mike is doing here. That should lead to a far-more reproducible build, when building with endpoint or in isolation. @fntlnz do you agree (if so can we get another +1)?

fntlnz (approved):

Issue #66, opened 2022-01-25
EventProbe:  Compatibility with Kernels >= 5.10


fntlnz:
We agreed that as a starting point, for the EventProbe (network, file and process) , we want to support kernels >= 5.10.

We need to make sure our usage of helpers and function arguments is compatible with this requirement.

fntlnz:
Regarding file events,https://github.com/elastic/ebpf/pull/65 contains  the changes to make them compatible with this requirement.

mmat11:
Closing this as we now have specific issues: https://github.com/elastic/ebpf/issues/91, https://github.com/elastic/ebpf/issues/92. Existing helpers have been adjusted already.

PR #77, opened 2022-03-08
Add setuid/setgid events

rhysre:
Adds setuid/setgid events. These are not required for any of the "wide events" that we need to build session view, but we need them for compatibility with the old kprobes/tracefs events implementation on Linux.

Note that we send up events on changed [user/group] real id, effective id, *saved id, fs id*. The last two (saved/fs) we don't currently send to ECS, they're not in the schema (saved is sent to ECS in the new wide effect model, fs isn't sent at all, old or new event model). 

The existing kprobes/tracefs implementation however, sends a uid/gid change event for every invocation of `proc_id_connector` when passed the `PROC_EVENT_UID` or `PROC_EVENT_GID` parameters. It's called [from here](https://elixir.bootlin.com/linux/v5.9.9/source/kernel/cred.c#L497) in `commit_creds`:

```c
	/* send notifications */
	if (!uid_eq(new->uid,   old->uid)  ||
	    !uid_eq(new->euid,  old->euid) ||
	    !uid_eq(new->suid,  old->suid) ||
	    !uid_eq(new->fsuid, old->fsuid))
		proc_id_connector(task, PROC_EVENT_UID);

	if (!gid_eq(new->gid,   old->gid)  ||
	    !gid_eq(new->egid,  old->egid) ||
	    !gid_eq(new->sgid,  old->sgid) ||
	    !gid_eq(new->fsgid, old->fsgid))
		proc_id_connector(task, PROC_EVENT_GID);
```

This will result in a uid/gid change event being sent for a saved/fs uid/gid, even though that's not reported in the ECS schema. I just duplicate this behaviour in this PR (and will in endpoint) to maintain the existing behaviour, but it's a good thing to be aware of. @nicholasberlin Can you confirm my understanding is correct.

`commit_creds` has been in the kernel unchanged for well over a decade, so it should be a fairly stable thing to attach a kprobe to.

### DIFF
--- a/GPL/EventProbe/Process/Probe.bpf.c
+++ b/GPL/EventProbe/Process/Probe.bpf.c
@@ -157,3 +157,47 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
 out:
     return 0;
 }
+
+SEC("fentry/commit_creds")
+int BPF_PROG(fentry__commit_creds, struct cred *new)
+{
+    const struct task_struct *task = (struct task_struct *)bpf_get_current_task_btf();
+    const struct cred *old         = task->real_cred;
+
+    // NB: We check for a changed fsuid/fsgid despite not sending it up.  This
+    // keeps this implementation in-line with the existing endpoint behaviour
+    // for the kprobes/tracefs events implementation.
+
+    if (new->uid.val != old->uid.val || new->euid.val != old->euid.val ||
+        new->suid.val != old->suid.val || new->fsuid.val != old->fsuid.val) {
+
+        struct ebpf_process_setuid_event *event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
+        if (!event)
+            goto out;
+
+        event->hdr.type = EBPF_EVENT_PROCESS_SETUID;
+        event->hdr.ts   = bpf_ktime_get_ns();
+
+        ebpf_pid_info__fill(&event->pids, task);
+        ebpf_cred_info__fill(&event->creds, task);
+        bpf_ringbuf_submit(event, 0);
+    }
+
+    if (new->gid.val != old->gid.val || new->egid.val != old->egid.val ||
+        new->sgid.val != old->sgid.val || new->fsgid.val != old->fsgid.val) {
+
+        struct ebpf_process_setuid_event *event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
+        if (!event)
+            goto out;
+
+        event->hdr.type = EBPF_EVENT_PROCESS_SETGID;
+        event->hdr.ts   = bpf_ktime_get_ns();
+
+        ebpf_pid_info__fill(&event->pids, task);
+        ebpf_cred_info__fill(&event->creds, task);
+        bpf_ringbuf_submit(event, 0);
+    }
+
+out:
+    return 0;
+}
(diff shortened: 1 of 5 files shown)


GPL/EventProbe/Process/Probe.bpf.c
```diff
@@ -157,3 +157,47 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
 out:
     return 0;
 }
+
+SEC("fentry/commit_creds")
+int BPF_PROG(fentry__commit_creds, struct cred *new)
+{
+    const struct task_struct *task = (struct task_struct *)bpf_get_current_task_btf();
```
mmat11, line 164: this helper is not available in 5.10, see https://github.com/elastic/ebpf/issues/66

GPL/EventProbe/Process/Probe.bpf.c
```diff
@@ -157,3 +157,47 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
...
+
+        ebpf_pid_info__fill(&event->pids, task);
+        ebpf_cred_info__fill(&event->creds, task);
+        bpf_ringbuf_submit(event, 0);
+    }
+
+    if (new->gid.val != old->gid.val || new->egid.val != old->egid.val ||
+        new->sgid.val != old->sgid.val || new->fsgid.val != old->fsgid.val) {
+
+        struct ebpf_process_setuid_event *event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
```
mmat11, line 189:
```suggestion
        struct ebpf_process_setgid_event *event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
```
typo

rhysre: 👍 Thanks

GPL/EventProbe/Process/Probe.bpf.c
rhysre: 👍  Thanks for the context. Removed it from here and also removed it from the setsid probe. 

I was wondering where all the `BPF_CORE_READ` invocations came from that had replaced the raw memory dereferences.

Makes me sad as I was really digging the raw kernel memory dereferences, alas, maybe in a few years when everybody is well past 5.10.

mmat11 (approved):
LGTM

GPL/EventProbe/Process/Probe.bpf.c
```diff
@@ -135,7 +135,7 @@ int BPF_PROG(fentry__taskstats_exit, const struct task_struct *task, int group_d
 SEC("tracepoint/syscalls/sys_exit_setsid")
 int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
 {
-    const struct task_struct *task = (struct task_struct *)bpf_get_current_task_btf();
+    const struct task_struct *task = (struct task_struct *)bpf_get_current_task();
```
nicholasberlin, line 138: why this change?

nicholasberlin (approved):
Nice work.

GPL/EventProbe/Process/Probe.bpf.c
rhysre: @mmat11 Pointed out that we shouldn't be using `bpf_get_current_task_btf` as it isn't available on 5.10, which we want to support. Removed the use in this PR and also removed this existing use in the setsid probe while I was at it.

Issue #79, opened 2022-03-14
[BUG] eBPF skeleton fails to load on kv 5.10 due to fentry__do_renameat2


lrishi:
do_renameat2 is static in 5.10
```
libbpf: failed to find kernel BTF type ID of 'do_renameat2'
libbpf: prog 'fentry__do_renameat2': failed to prepare load attributes: -3

```

PR #81, opened 2022-03-17
EventProbe: add the ability to disable progs autoload, add rename tra…

mmat11:
Closes https://github.com/elastic/ebpf/issues/79


non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -181,17 +179,49 @@ static int fill_ctx_relos(struct ebpf_event_ctx **ctx)
...
+/* Certain programs are available only if certain requirements are met.
+ *
+ * E.g. `do_renameat2` is not present in a 5.10 BTF, so tracepoints for the rename
+ * syscalls will be loaded instead.
+ */
+static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj)
+{
+    int err = 0;
+
+    if (BTF_FUNC_EXISTS(btf, do_renameat2)) {
```
nicholasberlin, line 194: this logic looks backward to me.

mmat11: mmm? autoload is being set to `false` so it should be correct the way it is.

CI logs on aarch64 5.10:
```
..."message":"libbpf: prog 'fentry__do_renameat2': skipped loading\n", ...
..."message":"libbpf: prog 'fexit__tcp_v6_connect': skipped loading\n", ...
```

nicholasberlin (approved):
LGTM, some comment suggestions.

GPL/EventProbe/File/Probe.bpf.c
```diff
@@ -39,24 +39,23 @@ DECL_FUNC_ARG_EXISTS(vfs_rename, rd);
 SEC("fentry/do_unlinkat")
 int BPF_PROG(fentry__do_unlinkat)
 {
-    struct ebpf_fileevents_state state;
-    __builtin_memset(&state, 0, sizeof(struct ebpf_fileevents_state));
-    ebpf_fileevents_state__set(EBPF_FILEEVENTS_STATE_UNLINK, &state);
+    struct ebpf_events_state state = {};
```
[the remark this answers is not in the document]
mmat11: to my understanding, `{}` creates a zero-initialized struct -> https://godbolt.org/z/h8b8c53PE

mmat11:
@nicholasberlin @lrishi @stanek-michal @rhysre this is ready for a review

non-GPL/LibEbpfEvents/LibEbpfEvents.h
```diff
@@ -27,13 +28,19 @@ typedef int (*ebpf_event_handler_fn)(struct ebpf_event_header *);
...
+ * returns 0 on success or less than 0 on failure.
+ *
+ * TODO: add `ebpf_event_ctx__load_attach` to API?
  */
 int ebpf_event_ctx__new(struct ebpf_event_ctx **ctx,
                         ebpf_event_handler_fn cb,
                         uint64_t features,
-                        uint64_t events);
+                        uint64_t events,
+                        bool poke_load_attach);
```
lrishi, line 43: I might be wrong, but it seems to me like the caller has to unnecessarily define a ctx var to be able to call this api, when poke_load_attach is set to true. Also, I dont see a null check for user supplied ctx ptr, the api uses it directly.

What do you think about getting rid of poke_load_attach and if ctx is NULL, treat it as a poke, essentially require the caller to specify a ctx location if they actually want to use functionality, else just mock and return success/failure. This should be fine as long as the behavior is documented in comments.

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -181,17 +179,52 @@ static int fill_ctx_relos(struct ebpf_event_ctx **ctx)
...
+                        bool poke_load_attach)
 {
     int err = 0;
-    *ctx    = calloc(1, sizeof(struct ebpf_event_ctx));
+
+    struct btf *btf = btf__load_vmlinux_btf();
+    if (libbpf_get_error(btf))
+        goto out_destroy_probe;
+
+    *ctx = calloc(1, sizeof(struct ebpf_event_ctx));
```
lrishi, line 227: Will be good to add a null check

lrishi (commented):
LGTM, few minor comments, if you think they need to be resolved, please do, else i'll approve.

non-GPL/LibEbpfEvents/LibEbpfEvents.h
mmat11: I like it!

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -191,33 +206,47 @@ int ebpf_event_ctx__new(struct ebpf_event_ctx **ctx,
...
     if (err != 0)
         goto out_destroy_probe;
 
-    err = EventProbe_bpf__attach((*ctx)->probe);
+    err = EventProbe_bpf__attach(probe);
     if (err != 0)
         goto out_destroy_probe;
 
+    if (!ctx)
+        goto out_destroy_probe;
```
lrishi, line 241: If we jump to clean here without ctx, we will not cleanup the probe and/or ringbuf, right? Needs few more lines I think?

mmat11: I am cleaning out the probe before cleaning ctx now, thanks

nicholasberlin (approved):

lrishi (approved):

PR #87, opened 2022-03-28
EventProbe: aarch64 support

mmat11:
1. finally use `features` to make decisions during probe load-link
2. add a small utility binary `LibEbpfEventsTestLoader` which tries to load the probe with every combination of features for a fast feedback during development
3. add the possibility to specify features in `EventsTrace`
4. adds aarch64 support

### DIFF
--- a/GPL/EventProbe/State.h
+++ b/GPL/EventProbe/State.h
@@ -24,7 +24,8 @@ enum ebpf_events_state_op {
     EBPF_EVENTS_STATE_UNKNOWN        = 0,
     EBPF_EVENTS_STATE_UNLINK         = 1,
     EBPF_EVENTS_STATE_RENAME         = 2,
-    EBPF_EVENTS_STATE_TCP_V6_CONNECT = 3,
+    EBPF_EVENTS_STATE_TCP_V4_CONNECT = 3,
+    EBPF_EVENTS_STATE_TCP_V6_CONNECT = 4,
 };
 
 struct ebpf_events_key {
@@ -48,7 +49,7 @@ struct ebpf_events_rename_state {
     struct vfsmount *mnt;
 };
 
-struct ebpf_events_tcp_v6_connect_state {
+struct ebpf_events_tcp_connect_state {
     struct sock *sk;
 };
 
@@ -56,7 +57,8 @@ struct ebpf_events_state {
     union {
         struct ebpf_events_unlink_state unlink;
         struct ebpf_events_rename_state rename;
-        struct ebpf_events_tcp_v6_connect_state tcp_v6_connect;
+        struct ebpf_events_tcp_connect_state tcp_v4_connect;
+        struct ebpf_events_tcp_connect_state tcp_v6_connect;
     };
 };
 
--- a/GPL/EventProbe/State.h
+++ b/GPL/EventProbe/State.h
@@ -34,6 +34,7 @@ struct ebpf_events_key {
 
 struct ebpf_events_unlink_state {
     struct vfsmount *mnt;
+    struct dentry de;
 };
 
 enum ebpf_events_rename_state_step {
--- a/non-GPL/LibEbpfEvents/LibEbpfEvents.c
+++ b/non-GPL/LibEbpfEvents/LibEbpfEvents.c
@@ -208,9 +208,6 @@ static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj
         err = err ?: bpf_program__set_autoload(obj->progs.kretprobe__do_filp_open, false);
         err = err ?: bpf_program__set_autoload(obj->progs.kprobe__vfs_rename, false);
         err = err ?: bpf_program__set_autoload(obj->progs.kretprobe__vfs_rename, false);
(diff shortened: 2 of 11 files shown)


lrishi:
lgtm, lets trigger a jenkins run for endpoint-dev right away and see if things go green.

non-GPL/LibEbpfEventsTestLoader/LibEbpfEventsTestLoader.c
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: Elastic-2.0
```
lrishi, line 1: This is super cool 👍

mmat11:
> lgtm, lets trigger a jenkins run for endpoint-dev right away and see if things go green.

endpoint needs to correctly set `features` now, it isn't ready to go as-is

non-GPL/LibEbpfEvents/LibEbpfEvents.h
```diff
@@ -14,10 +14,8 @@
 
 #include "EbpfEventProto.h"
 
-enum ebpf_kernel_features {
-    EBPF_KERNEL_FEATURE_BPF     = (1 << 0),
-    EBPF_KERNEL_FEATURE_RINGBUF = (1 << 1),
-    EBPF_KERNEL_FEATURE_BTF     = (1 << 2),
+enum ebpf_kernel_feature {
+    EBPF_FEATURE_BPF_TRAMP = (1 << 0),
```
lrishi, line 18: I'd like to understand more about the feature flags. I am getting confused about whether these are meant to be choice or support flags.

Do we want the caller to determine whats supported and then pass appropriate flags
OR
If the caller passes the flag, and feature is unsupported we ignore it.

Just trying to see how to not outsource the functionality of these flags to the caller.

mmat11: I don't know how this was originally meant to work, maybe @nicholasberlin / @rhysre can clarify this? I personally don't see much value in passing flags which would break the probe itself (eg. our probe wouldn't work on a relatively old kernel without ringbuf support so I don't see the point in having that specific flag)

GPL/EventProbe/State.h
```diff
@@ -34,6 +35,7 @@ struct ebpf_events_key {
 
 struct ebpf_events_unlink_state {
     struct vfsmount *mnt;
+    struct dentry de;
```
nicholasberlin, line 38: why not a pointer?

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -172,29 +173,64 @@ static int probe_fill_relos(struct btf *btf, struct EventProbe_bpf *obj)
...
+        err = err ?: bpf_program__set_autoload(obj->progs.kprobe__do_unlinkat, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kprobe__mnt_want_write, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kprobe__vfs_unlink, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kretprobe__vfs_unlink, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kretprobe__do_filp_open, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kprobe__vfs_rename, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kretprobe__vfs_rename, false);
         err = err ?: bpf_program__set_autoload(obj->progs.kprobe__tcp_v6_connect, false);
         err = err ?: bpf_program__set_autoload(obj->progs.kretprobe__tcp_v6_connect, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kprobe__do_renameat2, false);
```
nicholasberlin, line 213: What if `BTF_FUNC_EXISTS(btf, do_renameat2)` is `false`? Seems like both kprobe and ebpf will be disabled in  that case. Same question about tcp_v6_connect.

non-GPL/LibEbpfEvents/LibEbpfEvents.h
nicholasberlin: There was talk early on about where to drawn the support line, and these were here to allow for changes when things were decided. We decided to go with kernels that support ring buffers.

I think the idea was to use a [library](https://github.com/elastic/ebpf/pull/39) to figure out all the features that the kernel supports and pass that as the flags argument, and we would apply the best thing we could.

I don't the think the idea was fully fleshed out, perhaps Rhyse knows more or @tabell

non-GPL/LibEbpfEvents/LibEbpfEvents.c
mmat11: whoops, true

GPL/EventProbe/State.h
mmat11: I've tried several times, also tried inlining `vfs_unlink__exit`, but pathresolver would always break because of missing fields. Don't know the reason though

nicholasberlin (approved):
I'm curious about the dentry pointer not working in the saved state, but not gonna hold this up for it.

PR #88, opened 2022-03-29
contrib: update libbpf to v0.7.0, replace all deprecated functions and remove unneeded code

mmat11:
This will allow using new `libbpf_probe_*` functions


nicholasberlin (commented):
Code LGTM, I wonder if it will work on older kernels, like RHEL 8.

stanek-michal:
I'm looking at the new `libbpf` code and it seems the LINUX_VERSION number detection they did there is less robust than what we had in our client code. The way they did it, it seems targeted at Ubuntu primarily. I'll check if other old-kernel distros like RHEL8 work with the new feature detection. If they don't, it's better to keep our loop with different load methods to maintain wide support.

stanek-michal:
> I'm looking at the new `libbpf` code and it seems the LINUX_VERSION number detection they did there is less robust than what we had in our client code. The way they did it, it seems targeted at Ubuntu primarily. I'll check if other old-kernel distros like RHEL8 work with the new feature detection. If they don't, it's better to keep our loop with different load methods to maintain wide support.

I went through a few of our supported distros according to our [compatibility matrix](https://www.elastic.co/support/matrix), running host isolation gtests on each of them. I only tested x86_64 so far. RHEL8 and SLEL15 (SUSE) are unaffected by this PR. Debian 9 should be ignored because host isolation didn't work on it before due to lack of LPM_TRIE in the kernel. 
Debian 10, however, breaks because its actual LINUX_VERSION_CODE is different than what `uname` reports and the `/proc/version_signature` file that `libbpf` checks doesn't exist. 

I suspect Debian 10 could be the only one affected, as it was always the problem with Debian/Ubuntu distros that their `uname` output didn't match the actual LINUX_VERSION_CODE needed to load ebpf.

We should probably keep the old loading code with the multiple load methods to maintain the widest compatibility. That code (especially VDSO parsing) is a little hacky but it is only actually run in the rare occasion when the default method (no version override) fails.

mmat11:
> We should probably keep the old loading code with the multiple load methods to maintain the widest compatibility. That code (especially VDSO parsing) is a little hacky but it is only actually run in the rare occasion when the default method (no version override) fails.

I will revert the kernel version detection, would be nice if we could improve this upstream though.
Is there a method that _always_ works (looking at VDSO) so we can at least drop the "try in a loop" part?

nicholasberlin (approved):

nicholasberlin:
Oh, should this be updated to use new APIs:
https://github.com/elastic/ebpf/blob/8afc0bd03af992b481db938a0babf47753224265/non-GPL/LibEbpfEvents/LibEbpfEvents.c#L234-L239

mmat11:
> Oh, should this be updated to use new APIs:
> 
> https://github.com/elastic/ebpf/blob/8afc0bd03af992b481db938a0babf47753224265/non-GPL/LibEbpfEvents/LibEbpfEvents.c#L234-L239

will do in a follow up PR

stanek-michal (approved):
LGTM. Thanks for the changes.

PR #94, opened 2022-04-06
Initial check-in of multi-kernel testing functionality for BPF probes

rhysre:
This PR adds infrastructure to run tests on our BPF probes against a ton of kernels in the `testing/` directory. It's already caught a handful of bugs, and should hopefully continue to do so :)

[`testing/README.md`](https://github.com/elastic/ebpf/tree/rhysre/bpf-testing/testing#readme) documents its design/usage in detail, so I won't paste it in here, see that file. 

Few things in addition to that though:

- This whole setup is inspired by @florianl 's [bluebox](https://github.com/florianl/bluebox) project
- It currently only works on x86_64, but it shouldn't be that hard to extend it to aarch64, that's likely what I'm going to be working on when this is merged. 
- The testrunner is written in Go but I'm not a Go programmer. I've likely done a bunch of stuff in an un-golike way. @loshz @mrahs @Tacklebox @florianl  (and any other Go programmers) please rip my Go to shreds so this thing is as reliable as possible.
- If you run this locally, the container file operations tests will fail as our probes currently report file operations in a container relative to the root of the current mount namespace, when they should be breaking out (e.g. if a container accesses `/bin/ls` we should be reporting `/var/lib/docker/overlay2/<id>/merged/bin/ls` but currently we report simply `/bin/ls`). As per discussions with @m-sample, there is some ambiguity on what exactly we should be reporting here, so we'll need to have a discussion about that. In any case, those tests will be _very_ valuable when we've figured that out as detecting file operations on overlayfs with BPF is heavily subject to break between kernel releases.
- This runs fine locally, but I haven't given much thought as to how we're going to run it in CI. Theoretically we just need to spin up a beefy VM with KVM support and let it chew through the tests, but there may be some difficulties given the VM needs to be pretty beefy to run through every kernel in a reasonable amount of time.


testing/run_tests.sh
```diff
@@ -0,0 +1,224 @@
+#!/bin/bash
+
+SUMMARY_FILE="bpf-check-summary.txt"
+RESULTS_DIR="results"
+JOBS=$(nproc)
+
+build_initramfs() {
```
mmat11, line 7: can't this be done by [bluebox](https://github.com/florianl/bluebox) ?

testing/testrunner/utils.go
```diff
@@ -0,0 +1,177 @@
...
+	fmt.Println("")
+
+	PowerOff()
+}
+
+func AllTestsPassed() {
+	fmt.Println("ALL BPF TESTS PASSED")
+}
+
+func PowerOff() {
```
mmat11, line 159: shouldn't this be done by the caller? (`init`?)

testing/testrunner/main.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+func main() {
```
mmat11, line 3: this looks like a table driven test; can we use Go testing facilities to run these?

testing/testrunner/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/Elastic/Ebpf/testrunner
```
mmat11, line 1:
```suggestion
module github.com/elastic/ebpf/testrunner
```

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os/exec"
+	"time"
+)
+
+type EventsTraceInstance struct {
```
mmat11, line 11: what about having a `type Events []Event` which gets reset between every test and filled during the test?

`Event` will be an interface with `Type() EventType` and methods such as `func ProcessExec() ProcessExecEvent {...}`

this way we can test for total number of events per test and events ordering (doing this will require either some sort of timeout/deadline or setting an event cap after which assertions can start)

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,123 @@
...
+func (et *EventsTraceInstance) GetNextEventJson(types ...string) string {
+	var line string
+	for {
+		select {
+		case line = <-et.StdoutChan:
+			eventType := getJsonEventType(line)
+
+			for _, a := range types {
+				if a == eventType {
+					return line
```
loshz, line 79:
```suggestion
					break
```

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,123 @@
...
+				}
+			}
+		case <-time.After(60 * time.Second):
+			et.DumpStderr()
+			TestFail("timed out waiting for EventsTrace output, dumped stderr above")
+		}
+	}
+
+	return "" // Won't be hit
```
loshz, line 88:
```suggestion
	return line
```

testing/testrunner/tests.go
```diff
@@ -0,0 +1,291 @@
...
+		if err := json.Unmarshal([]byte(line), &setUidEvent); err != nil {
+			TestFail("failed to unmarshal JSON: ", err)
+		}
+
+		if setUidEvent.Pids.Tid == binOutput.PidInfo.Tid {
+			break
+		}
+	}
+
+	// TODO: Consult with @nicholasberlin on expected value of creds
```
loshz, line 186: Do we still need to do this?

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,123 @@
...
+func (et *EventsTraceInstance) Stop() error {
+	if err := et.Cmd.Process.Kill(); err != nil {
+		return err
+	}
+
+	if err := et.Cmd.Wait(); err != nil {
+		return err
+	}
+
+	return nil
```
loshz, line 100:
```suggestion
	return et.Cmd.Wait()
```

testing/testrunner/utils.go
```diff
@@ -0,0 +1,177 @@
...
+	testFuncName := runtime.FuncForPC(reflect.ValueOf(f).Pointer()).Name()
+
+	et := NewEventsTrace(args...)
+	et.Start()
+
+	f(et) // Will dump info and shutdown if test fails
+
+	fmt.Println("test passed: ", testFuncName)
+
+	et.Stop()
```
loshz, line 176: `Stop()` could return an error, do we want to check it?

testing/testrunner/utils.go
```diff
@@ -0,0 +1,177 @@
...
+
+func AssertPidInfoEqual(tpi TestPidInfo, pi PidInfo) {
+	AssertInt64Equal(pi.Tid, tpi.Tid)
+	AssertInt64Equal(pi.Tgid, tpi.Tgid)
+	AssertInt64Equal(pi.Ppid, tpi.Ppid)
+	AssertInt64Equal(pi.Pgid, tpi.Pgid)
+	AssertInt64Equal(pi.Sid, tpi.Sid)
+}
+
+func AssertStringsEqual(a string, b string) {
```
loshz, line 121: In Go, you can do this:
```suggestion
func AssertStringsEqual(a, b string) {
```

testing/testrunner/go.sum
```diff
@@ -0,0 +1,2 @@
+golang.org/x/sys v0.0.0-20220317061510-51cd9980dadf h1:Fm4IcnUL803i92qDlmB0obyHmosDrxZWxJL3gIeNqOw=
+golang.org/x/sys v0.0.0-20220317061510-51cd9980dadf/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
```
loshz, line 2: Remove this whole file.

loshz (commented):
Really great work, Rhys!

I've looked through the whole PR, but I'm going to focus on the Go code. It can be cleaned up a little, but I don't want to derail this PR with a bunch of nits.

A couple of improvements we could make in future iterations:
- Add CI for building/testing/linting.
- It might be more efficient to pass stdout/stderr bytes instead of a string to the respective channels.
- You should check for errors on the stdout/stderr scanners: https://pkg.go.dev/bufio#Scanner.Err

I'm more than happy to PR the above suggestions or pair on this with you!

testing/init/init.c
```diff
@@ -0,0 +1,148 @@
...
+
+    // Mount /proc and friends, everything needed for the testrunner to work
+    printf("[init] mounting pseudo-filesystems needed by tests\n", new_root);
+    CHECK(setup_pseudo_filesystems(), -1);
+
+    // Exec testrunner
+    printf("[init] exec'ing testrunner at /testrunner\n");
+    CHECK(execl("/testrunner", "/testrunner", NULL), -1);
+
+    return -1;
```
nicholasberlin, line 147:
```suggestion
    return 0;
```
right?

mmat11:
as discussed over Zoom/Slack, some changes to make use of the testing package https://github.com/elastic/ebpf/commit/f62fb01c258947bcce2d4fbd8ffe75bc3a8f037a

testing/run_tests.sh
rhysre: Yeah, it can. On an earlier iteration of this PR, I had bluebox checked in as a submodule and was using it directly. 

Ultimately I realized that bluebox's two primary functionalities of an initramfs builder and init program were better put in a bash script and C program respectively. It made the code simpler and allowed me to tailor things better to this use case.

testing/testrunner/utils.go
rhysre: The C `init` program performs an `execve` on `testrunner`, so it _is_ init / the only process running. It has to be the one to perform the poweroff.

testing/testrunner/go.mod
rhysre: 👍  done

testing/scripts/build_mainline_kernels.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
rhysre: done

testing/scripts/run_single_test.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
rhysre: done

testing/testrunner/tests.go
rhysre: This was something I forgot about, turned out I was just slightly confused. Resolved in the commit referencing UIDs/GIDs.

testing/testrunner/eventstrace.go
rhysre: done, let me know if that `break loop` statement is idiomatic go.

testing/testrunner/eventstrace.go
rhysre: done, see above

testing/scripts/run_single_test.sh
```diff
@@ -0,0 +1,53 @@
...
+SUCCESS_STRING="ALL BPF TESTS PASSED"
+
+KVM_ARGS=""
+if [[ -e /dev/kvm ]]
+then
+    KVM_ARGS="-enable-kvm -cpu host"
+fi
+
+qemu-system-x86_64 \
+    -nographic -m 1G \
```
florianl, line 35: Depending on the size of `initramfs.cpio` I recommend a larger value like `2G` or even `4G`.
As `initramfs.cpio` can be a compressed archive choosing a larger value for `-m` results in a significant speedup of the kernel boot time.

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,133 @@
...
+			et.StdoutChan <- scanner.Text()
+		}
+	}()
+
+	et.StderrChan = make(chan string, 100)
+	go func() {
+		scanner := bufio.NewScanner(et.Stderr)
+		for scanner.Scan() {
+			et.StderrChan <- scanner.Text()
+		}
```
florianl, line 49: It is good practice to check for scanner errors with `scanner.Err()`.

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,133 @@
...
+const eventsTraceBinPath = "/EventsTrace"
+
+func (et *EventsTraceInstance) Start() {
+	if err := et.Cmd.Start(); err != nil {
+		fmt.Println("failed to start EventsTrace: ", err)
+		TestFail()
+	}
+
+	et.StdoutChan = make(chan string)
+	go func() {
```
florianl, line 37: This go routine (and the following one below around line 45) are started without control and will exist longer than `EventsTraceInstance`.
To get control over these started Go routines, one could pass a `context.Context` as argument to start and check within the Go routines if this context is still correct or if the Go routine should exit.

This could also replace the `<-time.After(60 * time.Second)` with`<- ctx.Done()` and therefore replace the hard coded time value for all tests.

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,133 @@
...
+	Cmd        *exec.Cmd
+	Stdout     io.ReadCloser
+	Stderr     io.ReadCloser
+	StdoutChan chan string
+	StderrChan chan string
+}
+
+const eventsTraceBinPath = "/EventsTrace"
+
+func (et *EventsTraceInstance) Start() {
```
florianl, line 30: Maybe add a comment that this function is using a blocking element and will only return if the conditions (currently `<-et.StdoutChan` or `<-time.After(60 *time.Second)` are met.

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,133 @@
...
+func (et *EventsTraceInstance) DumpStderr() {
+	fmt.Println("===== EventsTrace Stderr =====")
+	for {
+		select {
+		case line, ok := <-et.StderrChan:
+			if !ok {
+				return
+			}
+			fmt.Println(line)
+		case <-time.After(1 * time.Second):
```
florianl, line 73: What is the intendend purpose of this hard coded time check?
Is the concern that `et.StderrChan` is blocking forever?

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,133 @@
...
+	var line string
+loop:
+	for {
+		select {
+		case line = <-et.StdoutChan:
+			eventType := getJsonEventType(line)
+
+			for _, a := range types {
+				if a == eventType {
+					break loop
```
florianl, line 89: Personal opinion: I think that is not a good place to use goto lables in Go.
```suggestion
					return line
```

testing/testrunner/utils.go
rhysre: Fixed.

testing/testrunner/go.sum
rhysre: done

testing/testrunner/utils.go
rhysre: 🙏 done

non-GPL/EventsTrace/CMakeLists.txt
```diff
@@ -6,7 +6,11 @@
 
 add_executable(EventsTrace EventsTrace.c)
 add_dependencies(EventsTrace LibEbpfEvents EventProbe_skeleton libbpf libelf)
-target_link_libraries(EventsTrace LibEbpfEvents ${LIBBPF_LIB} ${LIBELF_LIB} -lz)
+target_link_libraries(EventsTrace LibEbpfEvents ${LIBBPF_LIB} ${LIBELF_LIB} -lz -static)
+
+if (BUILD_STATIC_EVENTSTRACE)
+    target_link_libraries(EventsTrace -static)
```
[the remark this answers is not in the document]
rhysre: Oversight on my part from testing. Thanks for the catch Nick!

testing/README.md
```diff
@@ -0,0 +1,180 @@
...
+The `testrunner` binary is written in Go (see the `testrunner` directory). It
+starts
+[EventsTrace](https://github.com/elastic/ebpf/tree/main/non-GPL/EventsTrace)
+and runs various statically linked binaries passed into the VM by way of an
+[initramfs](https://www.kernel.org/doc/html/latest/filesystems/ramfs-rootfs-initramfs.html#what-is-initramfs)
+that generate events, which are then read by the testrunner and verified. This
+ensures we don't need to waste time starting up userspace so we can test
+against a large number of kernels quickly. See
+[bluebox](https://github.com/florianl/bluebox) for a more general
+implementation of this idea, that this checker was based on.
```
florianl, line 41: Tbf the code I see in `testing/run_tests.sh` and `/testing/init/` is a partial reimplementation of [bluebox](https://github.com/florianl/bluebox).
It misses the design goals of cross platform compatibility and lightweight dependency management.
In addition in [Debugging Kernels](https://github.com/elastic/ebpf/tree/rhysre/bpf-testing/testing#debugging-kernels) another tool ([virtme](git.kernel.org/pub/scm/utils/kernel/virtme/virtme)) is recommended for debugging.

So in total three different solutions are mentioned, one that provided the idea, one that is a partial implementation of that idea and one that is different from that. Overall this will result in a confusing user experience I think. The tool that is running the tests should also be used for debugging (maybe with additional arguments).

testing/init/init.c
rhysre: That line will never be hit, CHECK will either `return -1` or `execl` will succeed. I made it return -1 due to that, doesn't really matter though given the line will never be hit. I added a comment to clarify.

testing/scripts/run_single_test.sh
rhysre:
```
-rw-rw-r--  1 vagrant vagrant  31M Apr 11 21:34 initramfs.cpio
```

The initramfs.cpio I'm currently building is 31M. My understanding is that it's not compressed as it's currently being built in `run_tests.sh`, so 1G seems like plenty to me. I don't think we should ever be paging the contents of the initramfs to disk with 1G of memory. Am I missing something?

nicholasberlin (approved):

testing/testrunner/eventstrace.go
rhysre: done

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,138 @@
...
+	et.StdoutChan = make(chan string)
+	go func() {
+		scanner := bufio.NewScanner(et.Stdout)
+		for scanner.Scan() {
+			if err := scanner.Err(); err != nil {
+				fmt.Println("failed to read from EventsTrace stdout: ", err)
+				return
+			}
+			et.StdoutChan <- scanner.Text()
+		}
```
loshz, line 45: The error is only populated _after_ `scanner.Scan()` returns false. So you should check for errors afterwards:
```suggestion
		for scanner.Scan() {
			et.StdoutChan <- scanner.Text()
		}
		if err := scanner.Err(); err != nil {
			fmt.Println("failed to read from EventsTrace stdout: ", err)
			return
		}
```

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,138 @@
...
+	et.StderrChan = make(chan string, 100)
+	go func() {
+		scanner := bufio.NewScanner(et.Stderr)
+		for scanner.Scan() {
+			if err := scanner.Err(); err != nil {
+				fmt.Println("failed to read from EventsTrace stderr: ", err)
+				return
+			}
+			et.StderrChan <- scanner.Text()
+		}
```
loshz, line 57: The error is only populated after scanner.Scan() returns false. So you should check for errors afterwards:
```suggestion
		for scanner.Scan() {
			et.StderrChan <- scanner.Text()
		}
		if err := scanner.Err(); err != nil {
			fmt.Println("failed to read from EventsTrace stderr: ", err)
			return
		}
```

testing/testrunner/eventstrace.go
loshz: Agreed.

testing/testrunner/utils.go
```diff
@@ -0,0 +1,190 @@
...
+
+type ProcessForkEvent struct {
+	ParentPids PidInfo `json:"parent_pids"`
+	ChildPids  PidInfo `json:"child_pids"`
+}
+
+type ProcessExecEvent struct {
+	Pids     PidInfo  `json:"pids"`
+	Creds    CredInfo `json:"creds"`
+	Ctty     TtyInfo  `json:"creds"`
```
loshz, line 64: Struct field Ctty repeats json tag "creds":
```suggestion
	Creds    CredInfo `json:"creds"`
	Ctty     TtyInfo  `json:"ctty"`
```

loshz (changes requested):

testing/testrunner/eventstrace.go
rhysre: done

testing/testrunner/utils.go
rhysre: 🙏

testing/testrunner/eventstrace.go
rhysre: How would you re-write this? `return line` on line 99 necessitates a kinda wonky `return ""` at the end of the function that will never be reached, but needs to be there to satisfy the compiler? IMO this is slightly better.

testing/testrunner/eventstrace.go
rhysre: I took "no standard err has been output for one second" as signalling the end of stderr, but I really should have just closed the channel when stderr stopped instead to avoid the wait. Fixed.

testing/testrunner/eventstrace.go
rhysre: Done

testing/testrunner/eventstrace.go
rhysre: 👍 Done

testing/README.md
rhysre: In an early iteration of this work, I was using Bluebox directly, it was checked in at `testing/bluebox` and I was invoking it to build an initramfs and then run the tests with Qemu. After a few iterations, I came to the conclusion that it was simpler and cleaner to selectively re-implement parts of bluebox in a way specific to this use case.

The way I see it, conceptually, there are three major components to Bluebox:
- System setup logic (mount `/proc` etc.)
- Test running logic
- initramfs builder

IMO, for this use case, splitting out these three components to a C binary, a Go binary and a bash script, respectively was the cleanest way to do things. The system setup logic for this use case is pretty low level (`pivot_root` etc, see `init.c`). This is an application well-suited to C, but poorly suited to a higher-level language like Go. The test running logic is higher level, and thus well suited to a language like Go. Splitting these two things up therefore seemed like the best way forward.

After splitting those out, given building an initramfs in bash is super easy (`find <dir> | cpio --format=newc -ov > initramfs.cpio`), just moving that part to `run_tests.sh` seemed like the cleanest and simplest thing to do.

Regarding virtme, it's the best way to debug issues in misbehaving kernel images, and I don't think it will lead to a confusing user experience. You get a full shell in the kernel in question with access to all the tools of your host filesystem. Not totally sure what you had in mind when you say that the "tool that is running the tests should also be used for debugging", I don't think we can build a tool that's more powerful for debugging misbehaving kernel images than `virtme`, so IMO we might as well just recommend its use (if I'm reading you correctly here).

rhysre:
@mmat11 I've spent the last few hours looking at your table driven test code as well as thinking about the other points you brought up. Addressing them in this comment:

## 1) That we should use table-driven tests

I totally see where you're coming from with respect to reducing boilerplate in the test functions, if we could get rid of the loops in the test functions, that would be great, and is the type of thing a table driven test seems to be designed to eliminate. IMO however, we should keep the code as-is for this iteration.

As discussed, using the table-driven test idea makes it hard to filter by specific fields in an event coming out of EventsTrace. e.g. we need to get a fork event, but specifically the fork event with tid 1832. This is the type of work that's being done in the rather boilerplate-like `for` loops in the test functions.

We could avoid this by using reflection to do something like the following:

```go
processForkEvent := GetNextEventByTypeAndFields("PROCESS_FORK", map[string]interface{}{ "ParentPids.Tid", binOutput.Pid })
```

This is totally possible with the reflect package, and would remove most of the boilerplate in the loop of each test. IMO however, given the relative simplicity of the tests right now, it's overkill. When tests start getting more complicated and the amount of "event filtering" boilerplate really starts to grow, I think we can try implementing this approach, but right now, I think the code is fine as-is.

## 2) That we should use golang's built-in test framework instead of rolling our own

I don't think golang's built-in unit testing functionality is the right fit for this type of functional testing of an external program. Not a Go programmer (as stated), but it seems more geared towards unit testing Go code, instead of driving large functional tests of external stuff, correct me if I'm wrong, but this is the impression I get. 

Due to the specifics of our use case (running as init in a VM), I think running as a custom Go program where we control everything instead of leveraging Go's unit-testing functionality makes more sense. For example, on test end we need to `poweroff` instead of exit with status 0 for example, something that I think would be a bit wonky if we were using Go's testing framework, which is designed to just `exit` with a status indicating the test result.

## 3) That the framework should be runnable on the host system in addition to VMs

You made the point that the tester should be runnable outside a VM as that would provide a quick and convenient sanity-check for BPF changes.  Been thinking about that a fair bit today and I also drew the conclusion that we shouldn't do that, at least for this first iteration. 

I see where you're coming from, it's quicker and snapper to run tests on the host system. However, the additional code and complexity this introduces IMO isn't worth it for the utility you get. I think this is especially true given that most of the time you can just copy your host kernel from `/boot` and test with that instead. 

Some of the test_bins (the container tests especially) make assumptions that they're running in the VM environment (e.g. that they're running as root so they can mount stuff, and that the cwd is /). While we could totally rework these such that relevant parameters like "path of file/mount to create" are passed in via argv or something, again, the complexity this introduces IMO isn't worth it given you can just run the tests on your current kernel by copying the bzImage out of `/boot`, it will be slower, but totally useable.

===

Just my 2 cents on all the points you brought up. IMO the PR is good to merge as-is given the state of things right now. I think points 1 and 3 are definitely worth revisiting if the tests get more complicated, but for the time being, aren't necessary, especially given time constraints on getting this merged and up and running.

testing/testrunner/eventstrace.go
```diff
@@ -0,0 +1,123 @@
...
+	if err := et.Cmd.Wait(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func NewEventsTrace(args ...string) *EventsTraceInstance {
+	var et EventsTraceInstance
+	args = append(args, "--print-initialized", "--unbuffer-stdout", "--libbpf-verbose", "--set-bpf-tramp")
```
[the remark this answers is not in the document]
rhysre: Is there any reason that `EventsTrace` doesn't just always use autodetect? Why do the features need to be controllable from the command line? Shouldn't it just always autodetect?

testing/README.md
florianl: Thanks for the explanation of this turn.

With the partial reimplementation you did in C I see the following problems:
- cross platform compatibility
  Iirc a goal is/was to be able to run these tests on ARM64 for x86_64 and vice versa.
- dependency management 
  For developers and CI/CD dependencies should be transparent. To my knowledge `cpio` is not a standart tool installed by default with every distro. Does homebrew provide it?
- virtme
  Falling back to `virtme` in case of issues opens the door for heisenberg bugs. E.g. this custom reimplementation mounts elements in `/sys` differently. In addition `virtme` does not work for ARM64. What is the suggestion here?

mmat11:
> @mmat11 I've spent the last few hours looking at your table driven test code as well as thinking about the other points you brought up. Addressing them in this comment:
> snip

feel free to merge as-is, we can change/improve it later if we need to

testing/README.md
rhysre: > * cross platform compatibility
>   Iirc a goal is/was to be able to run these tests on ARM64 for x86_64 and vice versa.

This really shouldn't be that difficult. We just need to cross-compile the C test binaries and init program and then just invoke the aarch64 qemu machine instead of the x86_64 one. AFAICS this should boil down to a few changes to the bash scripts e.g:

```
$ sudo apt install gcc-10-aarch64-linux-gnu
[snip]
$ aarch64-linux-gnu-gcc-10 init/init.c
$ readelf -a a.out | grep 'Machine'
Machine:                           AArch64
```

> * dependency management
>   For developers and CI/CD dependencies should be transparent. To my knowledge `cpio` is not a standart tool installed by default with every distro. Does homebrew provide it?

`cpio` is pretty ubiquitous. It came out of Bell labs with v7 Unix back in the 70s originally and has been a common tool on most *nix systems since then. It's installed by default on my mac and my Ubuntu VM, I'd assume it's the same for most major Linux distros.

> * virtme
>   Falling back to `virtme` in case of issues opens the door for heisenberg bugs. E.g. this custom reimplementation mounts elements in `/sys` differently. In addition `virtme` does not work for ARM64. What is the suggestion here?

The environment virtme creates when you give it a kernel image is for sure different from the environment that the tests run in, but IMO should be close enough to provide a lot of utility. The underlying kernel will be the same, and ultimately, that's the component this system is meant to test against. For bugs that stem from kernel differences, which should be most of the bugs this tools finds, virtme should be a great tool (e.g. verifier errors).

What would you suggest as alternative? The way I see it, a debugging tool for issues this multi-kernel tester finds should just spin up the kernel in a VM and give you a shell, which is exactly what virtme does. I'm sure we could re-implement virtme such that it gives you a shell in an environment that's _exactly_ the same as what the tests run in, and works on ARM64, but IMO the added utility we get out of that really isn't worth the development time.

loshz (approved):
:100:

PR #97, opened 2022-04-08
Fix path resolver sometimes leaving the buffer containing garbage

rhysre:
The buf argument to the dentry resolver points into a ringbuffer, and
is thus uninitialized. If we don't zero it out, the buf[0] == '\0' check
at the end of the function might see a garbage, non-null byte in buf[0]
and assume a path has been constructed, leaving it as garbage. We need
to zero out the buffer.

This PR also fixes up parts of EventsTrace that were printing invalid JSON,
which was required to debug this.


non-GPL/EventsTrace/EventsTrace.c
```diff
@@ -149,7 +150,38 @@ static void out_int(const char *name, const long value)
 
 static void out_string(const char *name, const char *value)
 {
-    printf("\"%s\":\"%s\"", name, value);
+    printf("\"%s\":\"", name);
+    for (size_t i = 0; i < strlen(value); i++) {
+        char c = value[i];
+        switch (c) {
+        case '\n':
```
mmat11, line 157: aren't most of these part of `iscntrl()`?

non-GPL/EventsTrace/EventsTrace.c
```diff
@@ -149,7 +150,38 @@ static void out_int(const char *name, const long value)
...
+            printf("\"");
+            break;
+        case '\t':
+            printf("\\t");
+            break;
+        case '\b':
+            printf("\\b");
+        default:
+            printf("%c", c);
+            break;
```
mmat11, line 176: should this be:
```suggestion
            break;
        default:
```
?

GPL/EventProbe/PathResolver.h
```diff
@@ -76,6 +76,10 @@ ebpf_resolve_path_to_string(char *buf, struct path *path, const struct task_stru
     struct dentry *curr_dentry = BPF_CORE_READ(path, dentry);
     struct dentry **dentry_arr;
 
+    // Ensure we zero out buf early up here so if we exit with any sort of
+    // error, we won't leave garbage in it if it's uninitialized
+    memset(buf, '\0', sizeof(buf));
```
nicholasberlin, line 81: >  If we don't zero it out, the buf[0] == '\0' check at the end of the function might see a garbage

Which line are you referring to?

Further, `sizeof(buf)` is equivalent to `sizeof(char *)` right? I think this will only clear the first 64-bits of the buffer.

non-GPL/EventsTrace/EventsTrace.c
rhysre: Yeah they all are, we need to catch them in their own cases to print the "canonical" escape code for each one though. e.g. '\n' instead of '\x0a`. `jq` will fail if it doesn't see the canonical escape for newlines, tabs and other common escapes.

non-GPL/EventsTrace/EventsTrace.c
rhysre: 🙏 tyvm, I was far too tired Thursday evening 😛

mmat11 (approved):

GPL/EventProbe/PathResolver.h
rhysre: Thanks for always catching stuff like this Nick.

Yeah, this should be `buf[0] = '\0'`. I'm referring to the very last conditional in the function, this one:

```c
    // Special case: root directory. If the path is "/", the above loop will
    // not have run and thus path_string will be an empty string. We handle
    // that case here.
    if (buf[0] == '\0') {
        buf[0] = '/';
        buf[1] = '\0';
    }
```

The problem arises if the path being constructed is `/` and `buf[0]` contains some garbage non-null byte when the function is called. Then, that conditional doesn't get entered, leaving the garbage non-null byte in `buf[0]` (and potentially more garbage after it). That memset actually does fix things, since it nulls out the first byte of the buf, but since we don't have a `len` argument, we really should be doing `buf[0] = '\0'` instead. Fixed it.

GPL/EventProbe/PathResolver.h
```diff
@@ -76,6 +76,10 @@ ebpf_resolve_path_to_string(char *buf, struct path *path, const struct task_stru
     struct dentry *curr_dentry = BPF_CORE_READ(path, dentry);
     struct dentry **dentry_arr;
 
+    // Ensure we make buf and empty string early up here so if we exit with any
```
nicholasberlin, line 79:
```suggestion
    // Ensure we make buf an empty string early up here so if we exit with any
```

nicholasberlin (approved):

PR #103, opened 2022-06-03
EventProbe: capture process tty write

mmat11:
- [x] tests

Co-Authored-By: Richard Robbins <mail@rcr.io>

### DIFF
--- a/GPL/EventProbe/Process/Probe.bpf.c
+++ b/GPL/EventProbe/Process/Probe.bpf.c
@@ -16,6 +16,12 @@
 #include "Helpers.h"
 #include "PathResolver.h"
 
+/* tty_write */
+DECL_FUNC_ARG(tty_write, from);
+DECL_FUNC_ARG(tty_write, buf);
+DECL_FUNC_ARG(tty_write, count);
+DECL_FUNC_ARG_EXISTS(tty_write, from);
+
 SEC("tp_btf/sched_process_fork")
 int BPF_PROG(sched_process_fork, const struct task_struct *parent, const struct task_struct *child)
 {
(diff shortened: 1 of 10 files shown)


non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -297,6 +308,8 @@ int ebpf_event_ctx__new(struct ebpf_event_ctx **ctx,
     if (opts.features_autodetect)
         probe_set_features(&opts.features);
 
+    probe->rodata->endpoint_pid = getpid();
```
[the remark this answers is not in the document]
mmat11: yes, will work with any `ebpf_event_ctx__new()` caller. Naming: do you have any suggestion?
I couldn't think of a generic name

tabell:
hmm maybe consumer pid? self pid?

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -188,6 +190,13 @@ static int probe_fill_relos(struct btf *btf, struct EventProbe_bpf *obj)
     }
     err = err ?: FILL_FUNC_RET_IDX(obj, btf, vfs_rename);
 
+    if (FILL_FUNC_ARG_EXISTS(obj, btf, tty_write, from)) {
+        err = err ?: FILL_FUNC_ARG_IDX(obj, btf, tty_write, buf);
```
nicholasberlin, line 194: The feels backwards to me ... it reads to me like: if `from` is a parameter then set the parameter index for `buf` and `count`.

Am I reading it wrong?

GPL/EventProbe/Process/Probe.bpf.c
```diff
@@ -234,3 +240,76 @@ int BPF_KPROBE(kprobe__commit_creds, struct cred *new)
...
+
+    struct ebpf_process_tty_write_event *event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
+    if (!event)
+        goto out;
+
+    event->hdr.type          = EBPF_EVENT_PROCESS_TTY_WRITE;
+    event->hdr.ts            = bpf_ktime_get_ns();
+    event->tty_out_len       = count;
+    event->tty_out_truncated = count > TTY_OUT_MAX ? 1 : 0;
+    bpf_probe_read_user(event->tty_out, count > TTY_OUT_MAX ? TTY_OUT_MAX : count, (void *)buf);
```
nicholasberlin, line 260: Should we test the result of this read and NULL (or something else) out the buffer on failure?

More to the point, if this fails will the there be garbage in the buffer? Is the buffer cleared by `bpf_ringbuf_reserve`? (I assume not).

nicholasberlin:
Oh, I just saw this was in draft mode. Too quick on the draw with my review.

non-GPL/LibEbpfEvents/LibEbpfEvents.c
mmat11: this is counterintuitive given the function name, but the macro returns 0 if the parameter `from` exists so it should be correct as is. I wouldn't change the macro as it's consistent with the other macros and it should be read as `if (fill_***())` rather than `if (exists_***())`

GPL/EventProbe/Process/Probe.bpf.c
mmat11: I suppose I could reset count to `0` if this read fails. I wouldn't clear the buffer since we are passing the length (count) along with it

mmat11:
> Oh, I just saw this was in draft mode. Too quick on the draw with my review.

It's reviewable, I am just adding a test

non-GPL/LibEbpfEvents/LibEbpfEvents.c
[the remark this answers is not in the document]
mmat11: I substituted `endpoint` with `consumer` as suggested

GPL/EventProbe/Process/Probe.bpf.c
mmat11: If `bpf_probe_read_user` fails (which in normal cases shouldn't) I am now discarding the event directly and output a message via bpf_printk

nicholasberlin (approved):

lrishi (approved):
lgtm 👍🏽

nicholasberlin:
Something else ...

`vim` was routinely writing 1024-ish bytes to update the screen, which is lower than the default buffer size, but in the same neighborhood.

How much have we thought about the buffer size?

Alternatively, bash writes every character as it's typed, which leads to a lot of buffer waste.
Not sure what the right thing to do is, just wanted to highlight it.

mmat11:
@nicholasberlin not sure either about the buffer. We can keep it as is for now and change it if we realize most events don't utilize it

mmat11:
Merging this, I'll eventually add other data we need in a separate PR

PR #110, opened 2022-06-22
Adds external/supplementary user to exec events

Tacklebox:
This adds the value of an environment variable, currently
"K8S_USER" but subject to change, to exec events if exists. This
involves lots of string comparisons, so it eats up a lot of
instructions. The number of environment variables that are able to be
checked is currently limited to 100. The number of instructions is also
highly dependent on the name to be searched. So when deciding on the final
variable name to check, this must be taken into account.

If possible, the admission controller should try to set the variable as
one of the first vars. If we can lower the number of vars to check, we
can gain back some headroom to either use a more descriptive name or add
more functionality to the exec hook.

This PR is just for feedback and discussing implementation details and limitations. The requirements for the feature are not complete.


lrishi:
This feature has been deferred indefinitely. Closing for now.

PR #111, opened 2022-06-29
Revamp Testing Infrastructure and run Multi-Kernel Tests in CI

rhysre:
This PR revamps the multi-kernel test setup in line with concerns recently
brought up by obs-profiling, extends it to support arm64 tests, and sets the
whole thing up to run in CI via GitHub actions.

## Background

Concerns were brought up by obs-profiling about the initial iteration of the
testing framework locking them into something that wasn't adequate with regards
to cross platform compatibility.

Specifically, obs-profiling wanted the core of the framework to be very easy to
run on different architectures generally (be it generating a test initramfs for
x86 on an arm64 Mac or running a test in an arm64 guest on an x86 host).
Practically speaking, this means writing the core components of the framework
in Go, as it's super easy to cross compile for different architectures and
platforms.

While these concerns were less of a problem for AWP, obs-profiling still wanted
to be able to share common logic pertaining to running BPF tests with us as to
follow the DRY principle. The concern was that either obs-profiling would have
to create something else from scratch, thus performing redundant work, or use a
framework that wasn't quite what they were looking for.

Upon discussion with @florianl, the common logic that both teams will need is
really just limited to initramfs generation and test binary running (i.e. the
init process in the VM). Anything beyond that (e.g. specifics of the tests,
driver scripts to invoke QEMU) becomes pretty team specific.

As a result, we decided that both teams can just `go install` his
[Bluebox](github.com/florianl/bluebox) project to provide this core
functionality of initramfs generation and a VM init process. Bluebox is written
in Go and easily compiles to run on arm64/x86 hosts or generate initramfs's for
arm64/x86 guests. Anything beyond what Bluebox provides is team-specific and
can be implemented by the team writing the tests.

This will maximise the amount of shared code between our testing setups (which
is good from a DRY standpoint) while also allowing each team to build what
works best for them on top of Bluebox.

This PR provides the AWP-side realiziation of that work, and also sets up the
whole thing to run in CI via GitHub actions.

## Changes Made

All initramfs generation logic has been moved into Bluebox. Thus the `init/`
directory and parts of `run_tests.sh` that were responsible for invoking `cpio`
to generate an initramfs have been removed. That job is now done by Bluebox.

The `make container` make target has been revamped to build a container that
can cross-compile the repo for x86_64 and arm64. See
`docker/Dockerfile.builder`. Building the repo for different architectures is
now as simple as `make ARCH=x86_64` or `make ARCH=aarch64`. Debug builds are
analgous, the target name is just `build-debug`.

The `build_mainline_kernels.sh` script has been updated to build all arm64 and
x86_64 kernel images we need, and output kernel binaries with debug info, so
kernel code can be stepped through with QEMU in case of a gnarly BPF error (see
the debugging section in the updated `testing/README.md`). These binaries are
stored in the GCS bucket under the `debug_binaries` directory of each
architecture directory.

## CI

As mentioned, a GitHub actions workflow has been added to run tests on multiple
kernels. It pulls kernel images from a GCS bucket and runs tests against them.

For this initial iteration, only mainline kernels (built with the
`build_mainline_kernels.sh` script) are run. Distro-specific kernels can be
easily added to the test matrix later via an extra line in the yaml.

Kernel images are cached via the cache action and only re-downloaded from GCS
if the containing directory changes to save bandwith and GCS costs.

It takes ~45s per kernel to run the kernel tests in a GitHub actions runner.
x86_64 and arm64 tests are run in parallel, and each encompass 8 kernels
currently (v5.11 - v5.18), so it takes ~6m to run them all. Unfortunately the
GitHub actions runners don't support KVM, which would drastically speed this
up. In the future, we can look at a self-hosted runner with KVM support if we
want to be really aggressive about testing on many, many kernels.

Whether tests pass or fail, a summary of the run and individual results are
archived in GitHub actions (see the summary page). `results-mainline-{arch}`
contains the console output for the tests run on each kernel and
`run-summary-{arch}.txt` contains a summary of the pass/fail status of each
test on that particular architecture.

## Bugs

A handful of bugs were discovered and fixed to get the tests running on all
kernels (see commits). ~~You'll note only one test is still failing --
`TestFileRename` on v5.11/aarch64. I'm opening this PR to get feedback now
while I continue to try to fix that, as fixing it has become a bit of a time
sink.~~ (Thank you @mmat11 for helping me squash the last bug 🚀 )

CC: @norrietaylor @SeanHeelan @m-sample


GPL/EventProbe/PathResolver.h
```diff
@@ -238,7 +238,21 @@ static void ebpf_resolve_kernfs_node_to_string(char *buf, struct kernfs_node *kn
...
 static void ebpf_resolve_pids_ss_cgroup_path_to_string(char *buf, const struct task_struct *task)
 {
-    struct kernfs_node *kn = BPF_CORE_READ(task, cgroups, subsys[pids_cgrp_id], cgroup, kn);
+    /*
+     * Since pids_cgrp_id is an enum value, we need to get it at runtime as it
+     * can change kernel-to-kernel depending on the kconfig or possibly not be
+     * enabled at all.
+     */
+    int cgrp_id;
+    if (bpf_core_enum_value_exists(enum cgroup_subsys_id, pids_cgrp_id)) {
```
mmat11, line 247: when compiling with endpoint's toolchain, I get:
```
In file included from /home/matt/elastic/ebpf/GPL/EventProbe/EventProbe.bpf.c:24:
In file included from /home/matt/elastic/ebpf/GPL/EventProbe/File/Probe.bpf.c:17:
/home/matt/elastic/ebpf/GPL/EventProbe/PathResolver.h:247:9: error: use of unknown builtin '__builtin_preserve_enum_value' [-Wimplicit-function-declaration]
    if (bpf_core_enum_value_exists(enum cgroup_subsys_id, pids_cgrp_id)) {
        ^
/home/matt/elastic/ebpf/build/libbpf-prefix/src/libbpf-target/bpf/bpf_core_read.h:176:2: note: expanded from macro 'bpf_core_enum_value_exists'
        __builtin_preserve_enum_value(*(typeof(enum_type) *)enum_value, BPF_ENUMVAL_EXISTS)
        ^
```

error doesn't occur with my system compiler (clang 14). From what I can see here https://github.com/llvm/llvm-project/blob/llvmorg-12.0.0/clang/test/Sema/builtins-bpf.c, we need at least v12+ (toolchain is v11)

mmat11: nvm, I had an old toolchain

nicholasberlin (commented):
Nice work. A couple of comments.

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -263,15 +264,35 @@ static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj
...
+     *
+     * While it may seem cleaner than an arch check, note that we cannot use
+     * libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL) here as it will
+     * return true on both aarch64 and x86_64. That function detects if a
+     * program type can be _loaded_ on the current machine.
+     * BPF_PROG_TYPE_TRACING can be loaded on aarch64, it will just fail to
+     * _attach_ (which that function doesn't indicate).
+     */
+    *features = 0;
+    if (strcmp(buf.machine, "x86_64") == 0)
```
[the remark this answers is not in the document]
mmat11: as an alternative: what about changing this probe https://github.com/libbpf/libbpf/blob/e99f34e14447866bcbc38944387531c36f6c85fe/src/libbpf_probes.c#L152 to do `bpf_prog_load` + attach?

non-GPL/LibEbpfEvents/LibEbpfEvents.c
```diff
@@ -263,15 +264,35 @@ static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj
...
+     * updated when that is merged into the mainline.
+     *
+     * While it may seem cleaner than an arch check, note that we cannot use
+     * libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL) here as it will
+     * return true on both aarch64 and x86_64. That function detects if a
+     * program type can be _loaded_ on the current machine.
+     * BPF_PROG_TYPE_TRACING can be loaded on aarch64, it will just fail to
+     * _attach_ (which that function doesn't indicate).
+     */
+    *features = 0;
```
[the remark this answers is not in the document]
mmat11: > I wasn't sure why this wasn't being done before

no reason, since we only have one feature it wouldn't make much sense to hardcode something and then autodetect it

non-GPL/LibEbpfEvents/LibEbpfEvents.c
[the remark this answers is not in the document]
rhysre: > Should we still libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL) in case the x64 kernel doesn't support it?

Good point, slipped my mind since AFAIK it's pretty ubiquitously enabled on x86 kernels. We should check anyways though.

To implement `kernel_supports_bpf_tramp`, we're going to have to load _and attach_ to detect support though (what Mattia said), which libbpf doesn't provide a feature probing function for.

`libbpf` pulls this hack where they pass a dummy `attach_btf_id` and expect `-EINVAL` in `libbpf_probe_bpf_prog_type` in order to detect if you can _load_ the program ([see here](https://github.com/libbpf/libbpf/blob/e99f34e14447866bcbc38944387531c36f6c85fe/src/libbpf_probes.c#L94)). We need to instead pass a valid `attach_btf_id` so we have a loaded dummy program and then try to attach it to something (thus detecting bpf trampoline support).

This seems tricky, although I need to dig into the low-level BPF syscall API to determine precisely how tricky. If it becomes a rabbit hole, it seems that BPF trampolines are enabled on x86 if `CONFIG_BPF_JIT` is enabled (then [this function](https://elixir.bootlin.com/linux/v5.11/source/arch/x86/net/bpf_jit_comp.c#L1738), is compiled into the kernel and used instead of [this one](https://elixir.bootlin.com/linux/v5.11/source/kernel/bpf/trampoline.c#L437)). Perhaps we can just get the value of that Kconfig to detect support instead? I'll investigate more on Tuesday.

testing/README.md
```diff
@@ -175,6 +122,72 @@ CONFIG_SCSI_VIRTIO
...
-the kernel you're trying to debug if it isn't working in virtme.
+kernels built with it should work fine in virtme. Distro kernels won't
+necessarily. Check the `.config` for the kernel you're trying to debug if it
+isn't working in virtme.
+
+## Debugging a Test Kernel Image With GDB
+
+Occasionally it's useful to attach a debugger to the kernel itself, to
+determine the exact cause of a particular BPF failure. To do this, you'll need
+a kernel _ELF binary_ (not a compressed image), _with_ debugging symbols. The
```
florianl, line 133: `_with_ debugging symbols` this is optional. Kernel sources also can be loaded withing gdb with `source`.

testing/README.md
```diff
@@ -175,6 +122,72 @@ CONFIG_SCSI_VIRTIO
...
+```
+
+Then, in another terminal, run `gdb` on your kernel ELF binary with debug
+symbols (usually called `vmlinux`). Connect to QEMU with `target remote
+localhost:1234` and  set your source search path to a locally-checked-out clone
+of Linux that matches the kernel you're debugging with `dir
+<path_to_linux_source>`.
+
+At this point, you'll be able to set breakpoints and continue execution with
+`c`. For example, to break at `start_kernel` (the main architecture-independent
```
florianl, line 163: I think it would be worth to name one or more symbols here, that are focused on debugging eBPF related issues - e.g. a symbol from one of the binaries under test.
there is a long way to go from `start_kernel` to calling `init` and then using a eBPF syscall.

testing/run_tests.sh
```diff
@@ -187,26 +106,18 @@ check_kvm() {
 
 usage() {
     echo "Usage: ./run_tests.sh [-m] [-d gs_bucket] [-j jobs]"
```
florianl, line 108: This option just got removed.
```suggestion
    echo "Usage: ./run_tests.sh  [-d gs_bucket] [-j jobs]"
```

non-GPL/LibEbpfEvents/LibEbpfEvents.c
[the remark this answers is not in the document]
rhysre: Ok I think I've tackled this, see latest commit message and comments within (it's somewhat involved so I won't repeat myself here).

I added a test for correct BPF tramp feature probing too, which is passing in CI.

.github/workflows/multikernel-tester.yml
```diff
@@ -0,0 +1,82 @@
...
+      uses: 'google-github-actions/setup-gcloud@v0'
+    - name: Create kernel images directory
+      run: |
+        sudo mkdir -p /kernel-images
+        sudo chown -fR $USER:$USER /kernel-images
+    # We get the hash of every kernel image and then hash the result. This results in a re-fetch
+    # if anything in the GCS directory changes at all
+    - name: Calculate kernel image cache key
+      run: |
+        echo "GCS_HASH=$(gsutil hash -h $HASH_PATH | md5sum | awk '{ print $1 }')" >> $GITHUB_ENV
```
lrishi, line 33: Rhys, we need to tweak this. This is what we want to accomplish:
1. Always restore from GH cache
2. gsutil rsync (which automatically takes care of the differences)
3. Restore back.

The way to do it is:
```
    - name: Get current time
      run: echo "TIMESTAMP=$(date +%s)" >> $GITHUB_ENV
    - uses: actions/checkout@v2
    - name: Setup GCS cache
      uses: actions/cache@v3
      with:
        key: gcs-cache-${{env.TIMESTAMP}}
        path: <local path>
        restore-keys: gcs-cache-
```
By giving the partial restore key, we restore from the latest cache, and since the key is always new, it gets recached with differences.

.github/workflows/formatting-build.yml
```diff
@@ -0,0 +1,47 @@
...
+      - main
+  pull_request:
+
+jobs:
+  formatting-and-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Test code formatting
+        run: make test-format
```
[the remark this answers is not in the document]
rhysre: Yeah, noticed this too. This is down to differences in clang-format versions. I've dockerized the `format` and `test-format` steps, so everyone will always use the same version (see latest commits). This should be fixed now.

testing/scripts/run_single_test.sh
```diff
@@ -1,13 +1,15 @@
...
+set -x
+
 # Copyright 2022 Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under
 # one or more contributor license agreements. Licensed under the Elastic
 # License 2.0; you may not use this file except in compliance with the Elastic
 # License 2.0.
 
 # Runs a BPF test on a single kernel.
-# Usage ./run_single_test.sh <kernel bzImage> <kernel name> <results file>
+# Usage ./run_single_test.sh <kernel bzImage> <kernel initramfs> <kernel name> <results file>
```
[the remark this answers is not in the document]
rhysre: Thanks, fixed.

testing/scripts/build_mainline_kernels.sh
```diff
@@ -64,11 +49,133 @@ do
...
+fi
+
+if [[ $# -ne 1 ]]
+then
+    echo "Usage ${0} <kernel source dir>"
+    exit 1
+fi
+
+mkdir -p chroot
+sudo debootstrap --arch amd64 --variant=minbase bullseye chroot http://ftp.us.debian.org/debian/
```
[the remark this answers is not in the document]
rhysre: This was done so that no setup on the host system (install packages, etc.) is required to build a fresh set of kernels. If we build a totally reproducible debian chroot and use that instead, everyone is using the same packages / tools / versions, changes in which can break the kernel build (e.g. see note in this script about pahole 1.22 being required).

I've added a comment to the script to explain this.

testing/README.md
rhysre: I don't think I fully follow here. If there's no DWARF information in the kernel ELF binary, the debugger won't know what instruction corresponds to which line of code and you won't be able to step through the source in the debugger. Using `source <file>` in GDB just runs the gdb commands in the given file AFAICS. Can you clarify?

.github/workflows/multikernel-tester.yml
```diff
@@ -0,0 +1,82 @@
...
+      env:
+        HASH_PATH: gs://ebpf-ci-kernel-images/${{ matrix.kernel_flavor }}/${{ matrix.arch }}/images/*
```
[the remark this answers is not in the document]
rhysre: I'd lean towards Florian's viewpoint here. While what we currently have _works_, since this is a public repository, it would be cleaner IMO to have all external artifacts required to test/build it publicly accessible.

I'm planning on brushing up a lot of the docs and removing some of the cruft in here shortly to make the repo more widely accessible to a public audience. In that regard it would be nice to make the bucket publicly available. Is that feasible @lrishi? Are there security / rate limit concerns with opening it up to the public?

I discussed basically this exact same issue with @tabell in the previous iteration of this but we didn't get to a final conclusion IIRC. Would be nice to do so and just make it public if possible.

docker/Dockerfile.builder
```diff
@@ -1,29 +1,45 @@
-ARG PULL_TAG=jammy
+FROM ubuntu:jammy
 
-FROM ubuntu:${PULL_TAG}
+ENV BPFTOOL_VERSION=5.15.0-33
```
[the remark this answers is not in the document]
rhysre: This should be a kernel version number. This was done because `bpftool` on Ubuntu is found in the `linux-tools-<kernel version number>` package ([see here](https://packages.ubuntu.com/jammy/devel/linux-tools-5.15.0-33)), where several such packages exist for different kernel versions.

The `linux-tools-<ver>` package contains a bunch of tools, some of which are version specific, hence the tying of the package to a specific kernel. `bpftool` should work fine for what we're using it for (skeleton generation) when compiled from newer kernel sources however, so we can just install a new one and run `update-alternatives` as done below.

On that note, `BPFTOOL_VERSION` is a confusing name for this envvar, changed to `LINUX_TOOLS_VERSION`.

testing/README.md
```diff
@@ -175,6 +122,72 @@ CONFIG_SCSI_VIRTIO
...
+is required so that function addresses line up with the data in the binary).
+
+On x86_64, for example, the invocation is:
+
+```
+qemu-system-x86_64 \
+    -s -S \
+    -nographic -m 1G \
+    -kernel <kernel image> \
+    -initrd <initramfs cpio archive> \
```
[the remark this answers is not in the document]
rhysre: 👍 good suggestion, done.

testing/README.md
rhysre: Changed it to `bpf_prog_load` instead.

Just noticed your suggestion to use a userspace symbol as well, that's a good call, I'll add that too when I get back to this on Monday.

docker/Dockerfile.builder
```diff
@@ -1,29 +1,46 @@
...
+    && echo "deb [arch=arm64] http://ports.ubuntu.com/ jammy multiverse" >> /etc/apt/sources.list.d/arm-sources.list \
+    && echo "deb [arch=arm64] http://ports.ubuntu.com/ jammy-updates multiverse" >> /etc/apt/sources.list.d/arm-sources.list \
+    && echo "deb [arch=arm64] http://ports.ubuntu.com/ jammy-backports main restricted universe multiverse" >> /etc/apt/sources.list.d/arm-sources.list \
+    && dpkg --add-architecture arm64 \
+    && sed -i 's/deb/deb [arch=amd64]/g' /etc/apt/sources.list \
+    && apt-get update
 
-RUN DEBIAN_FRONTEND=noninteractive apt-get update -y \
-    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+RUN apt-get update -y && \
```
lrishi, line 28: duplicate apt-get update

.github/workflows/multikernel-tester.yml
rhysre: Done, lmk if that's what you had in mind.

Makefile
```diff
@@ -1,39 +1,72 @@
...
-PWD ?= $(shell pwd)
-CONTAINER_RUNTIME ?= docker
+PWD = $(shell pwd)
 DOCKER_IMG_UBUNTU_VERSION ?= jammy
-BUILDER_PULL_TAG ?= us-docker.pkg.dev/elastic-security-dev/ebpf-public/builder:20220620-0715
+BUILDER_PULL_TAG ?= us-docker.pkg.dev/elastic-security-dev/ebpf-public/builder:20220707-1836
 BUILDER_TAG ?= us-docker.pkg.dev/elastic-security-dev/ebpf-public/builder:${USER}-latest
-C_INCLUDE_PATH ?=
-DOCKER_CACHE ?=
+CMAKE_FLAGS = -DARCH=${ARCH} -DBUILD_STATIC_EVENTSTRACE=True -DUSE_BUILTIN_VMLINUX=True -B${BUILD_DIR} -S${PWD}
```
lrishi, line 7: If I set CMAKE_FLAGS on shell, wiil it break the build or not take effect?

rhysre: It just won't take effect. I purposely changed most of the `?=`'s to `=`'s for this reason. IMO using `?=` for everything encapsulates the build logic in the Makefile to a much lesser extent and for that reason isn't quite as clean. The only thing remaining using `?=` is `ARCH`, which provides a nice simple "API" for building from the command line, `make ARCH=<arch>`, all other logic is encapsulated in the Makefile.

Feel free to disagree, this is mainly a matter of taste.

.github/workflows/multikernel-tester.yml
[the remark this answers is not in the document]
florianl: The problem here for me is, if there is an issue - e.g. the CI tests fail. I have no chance to debug it, provide feedback to questions of help.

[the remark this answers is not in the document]
rhysre: @florianl I've granted read access to the bucket to all Elastic employees pending further discussion on how we're going to open it up publicly. Let me know if you keep getting a 403.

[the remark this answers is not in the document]
florianl: thanks! it works now.

florianl (approved):
Thanks for the update! :pray:

GPL/EventProbe/PathResolver.h
```diff
@@ -238,7 +238,21 @@ static void ebpf_resolve_kernfs_node_to_string(char *buf, struct kernfs_node *kn
...
+        cgrp_id = bpf_core_enum_value(enum cgroup_subsys_id, pids_cgrp_id);
+    } else {
+        /* Pids cgroup is not enabled on this kernel */
```
mmat11, line 250: could you add a log message here?

mmat11 (approved):

testing/README.md
lrishi: @florianl Just wanted to confirm, if this  is alright, or we need to make a change here?

lrishi (approved):
lgtm 👍🏽 Great work Rhys!! Thanks for all the collab Florian!

testing/README.md
lrishi: Had a chat with @florianl offline about this, @rhysre Can you please discuss this offline and capture the details of alternate method suggested in the comment? This is the issue to track it: https://github.com/elastic/ebpf/issues/112
For now, I'll close this comment, thanks!

Issue #69, opened 2022-02-01
Agree on the Kernel version used to generate vmlinux.h


mmat11:
From https://github.com/elastic/ebpf/pull/67#discussion_r795800871

> We can probably just generate it once but we would need to agree on a kernel to be used as base, document that and when we need to update we need to agree on the newer kernel.

The idea is to always use the prebuilt vmlinux (https://github.com/elastic/ebpf/tree/main/contrib/vmlinux#vmlinux) to build `elastic/ebpf` both in CI and locally

fntlnz:
@leodido @rhysre @stanek-michal @nicholasberlin wdyt?

rhysre:
Correct me if I'm wrong, but isn't the best choice here to just generate one on 5.16 (current latest) and just use that so we have as many definitions as possible? It should be backwards compatible.

mmat11:
> isn't the best choice here to just generate one on 5.16 (current latest)

Yes, I think so

stanek-michal:
> Correct me if I'm wrong, but isn't the best choice here to just generate one on 5.16 (current latest) and just use that so we have as many definitions as possible? It should be backwards compatible.

In case some definitions/offsets changed in newer versions, my understanding is that BTF will resolve such incompatibilities anyway? In that case it seems to be a good idea to just use the latest version for vmlinux.h. But if we care about pre-BTF kernels in any scenario, we need to be careful.

rhysre:
> > Correct me if I'm wrong, but isn't the best choice here to just generate one on 5.16 (current latest) and just use that so we have as many definitions as possible? It should be backwards compatible.
> 
> In case some definitions/offsets changed in newer versions, my understanding is that BTF will resolve such incompatibilities anyway? In that case it seems to be a good idea to just use the latest version for vmlinux.h. But if we care about pre-BTF kernels in any scenario, we need to be careful.

Field offset changes are automatically handled by BTF relocations, if a field is actually _renamed_ or moved to a different struct, that detail will have to bleed into our code (e.g. via [bpf_core_field_exists](https://nakryiko.com/posts/bpf-core-reference-guide/#bpf-core-field-exists)) but the version of vmlinux.h we're using is totally independent of that and shouldn't affect things in any way.

fntlnz:
Agree on keeping our vmlinux in sync with the latest stable (now 5.16.5)

rhysre:
Closing this stale ticket as it's been resolved for a while.

PR #114, opened 2022-07-27
Cleanup CMake usage

rhysre:
This commit cleans up the cmake in this repository, removing a lot of
vestigial stuff from when this repo was just host isolation and putting
it in a state where it can be more cleanly built with endpoint.

Changes:

- Remove all the TARGET_* variables in the top level CMakelists.txt,
  push all that logic down to lower subdirectories
- Create a "libbpf" and "libelf" library target, instead of linking
  against `${LIBBPF_LIB}` and `${LIBELF_LIB}` and then manually including
  the required includes. This puts logic pertaining to where to find
  libbpf includes in cmake/libbpf.cmake instead of GPL/non-GPL, which
  means that we don't have to manually add the libbpf include
  directories and link libraries to every target that links against it.
- Remove a bunch of unneeded comments, message(...) invocations etc.
- Clean up the variables in cmake/libbpf.cmake and document why we need
  the one (LIBBPF_UAPI_INCLUDE_DIR) that remains
- Set ARCH_TRUNC based on ARCH, don't require a higher level project to
  manually specify it

Note that to get this PR building with endpoint, one LLVM intrinsic has been commented out as the endpoint toolchain
is currently using LLVM 11, which doesn't support `bpf_core_enum_value_exists` (see comment). This has been done to get reviews in early and will be removed when the toolchain is updated to LLVM 14. _This PR should *not* be merged until the toolchain is updated and the commented-out code has been removed_.

### DIFF
--- a/GPL/Events/PathResolver.h
+++ b/GPL/Events/PathResolver.h
@@ -243,22 +243,16 @@ static void ebpf_resolve_pids_ss_cgroup_path_to_string(char *buf, const struct t
      * can change kernel-to-kernel depending on the kconfig or possibly not be
      * enabled at all.
      */
-    // NOTE: This code is commented out to get this PR building and in a
-    // reviewable state as our toolchain currently uses LLVM 12, which doesn't
-    // support bpf_core_enum_value_exists. @stanek-michal is currently working
-    // on updating to LLVM 14. This will be removed when the toolchain is
-    // updated.
-    //
-    //    int cgrp_id;
-    //    if (bpf_core_enum_value_exists(enum cgroup_subsys_id, pids_cgrp_id)) {
-    //        cgrp_id = bpf_core_enum_value(enum cgroup_subsys_id, pids_cgrp_id);
-    //    } else {
-    //        /* Pids cgroup is not enabled on this kernel */
-    //        buf[0] = '\0';
-    //        return;
-    //    }
+    int cgrp_id;
+    if (bpf_core_enum_value_exists(enum cgroup_subsys_id, pids_cgrp_id)) {
+        cgrp_id = bpf_core_enum_value(enum cgroup_subsys_id, pids_cgrp_id);
+    } else {
+        /* Pids cgroup is not enabled on this kernel */
+        buf[0] = '\0';
+        return;
+    }
 
-    struct kernfs_node *kn = BPF_CORE_READ(task, cgroups, subsys[pids_cgrp_id], cgroup, kn);
+    struct kernfs_node *kn = BPF_CORE_READ(task, cgroups, subsys[cgrp_id], cgroup, kn);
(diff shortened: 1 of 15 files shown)


GPL/Events/PathResolver.h
```diff
@@ -243,16 +243,22 @@ static void ebpf_resolve_pids_ss_cgroup_path_to_string(char *buf, const struct t
...
-    if (bpf_core_enum_value_exists(enum cgroup_subsys_id, pids_cgrp_id)) {
-        cgrp_id = bpf_core_enum_value(enum cgroup_subsys_id, pids_cgrp_id);
-    } else {
-        /* Pids cgroup is not enabled on this kernel */
-        buf[0] = '\0';
-        return;
-    }
-
-    struct kernfs_node *kn = BPF_CORE_READ(task, cgroups, subsys[cgrp_id], cgroup, kn);
+    // NOTE: This code is commented out to get this PR building and in a
```
mmat11, line 246: maybe it's better to wait for the toolchain to be upgraded before merging this?

cmake/modules/BPF.cmake
```diff
@@ -11,7 +11,7 @@ set(LLC "llc")
 set(LLVM_STRIP "llvm-strip")
 set(BPFTOOL "bpftool")
 set(BTF_FILE "/sys/kernel/btf/vmlinux")
-option(USE_BUILTIN_VMLINUX "Whether or not to use the builtin vmlinux.h for building the BPF programs instead of trying to generate one from the system" True)
+option(USE_BUILTIN_VMLINUX "If true, use the builtin vmlinux.h for building BPF probes instead of generating one from system BTF" True)
```
mmat11, line 14: nit:
```suggestion
option(USE_BUILTIN_VMLINUX "If true, use the builtin vmlinux.h for building eBPF probes instead of generating one from system BTF" True)
```

also, not in scope for this PR, but IIRC we decided to always use the builtin vmlinux.h (https://github.com/elastic/ebpf/issues/69), right?

mmat11 (approved):
LGTM if it compiles

GPL/Events/PathResolver.h
rhysre: 👍 Yeah, this won't be merged until the toolchain is updated. 

I was just blocked waiting for the update so I did this to make things compile and get reviews while I'm waiting for the update to go through.

stanek-michal (approved):
This is great, thanks for the changes. I like the added comments about build quirks on older systems.

GPL/Events/PathResolver.h
stanek-michal: I'd say we can merge as it won't actually affect Endpoint until Endpoint updates the Ebpf submodule reference. The new toolchain will be ready soon and the switch will happen simultaneously with the update of submodule reference.

PR #117, opened 2022-08-12
Package kernel images in container images

lrishi:


nicholasberlin (approved):

docker/Dockerfile.imagepack
```diff
@@ -0,0 +1,6 @@
+FROM docker.io/ubuntu:latest
+COPY / /kernel-img-repository/
+
+LABEL org.opencontainers.image.source https://github.co/elastic/ebpf
```
mjwolf, line 4:
```suggestion
LABEL org.opencontainers.image.source https://github.com/elastic/ebpf
```

mjwolf (approved):
LGTM, other than one missing character 😀

docker/Dockerfile.imagepack
lrishi: 🙏🏽 Thanks

PR #123, opened 2022-08-18
Add Ubuntu kernels to multikernel testing

lrishi:


Makefile
```diff
@@ -135,22 +136,30 @@ endif
 # Update kernel images from gcs
 update-kims:
 	@mkdir -p LargeFiles
-	gsutil -m rsync -r gs://ebpf-ci-kernel-images LargeFiles
+	gsutil -m rsync -d -r gs://ebpf-ci-kernel-images LargeFiles
+
+# Build Containerized Kernel Image Pack(s)
+build-kips:
+	for dir in $(shell ls ./LargeFiles); do \
+	${CONTAINER_ENGINE} build -t ${IMAGEPACK_REPOSITORY}:$${dir}-${CURRENT_DATE_TAG} -f docker/Dockerfile.imagepack --build-arg IMGPACK_FILTER=$${dir} ./LargeFiles/$${dir}; \
```
nicholasberlin, line 144: I notice that `--no-cache` has been removed. 🤷

Makefile
```diff
@@ -135,22 +136,30 @@ endif
...
+	done
 
-# Build Containerized Kernel Image Pack
-build-kip:
-	${CONTAINER_ENGINE} build --no-cache -t ${IMAGEPACK_REPOSITORY}:${CURRENT_DATE_TAG} -f docker/Dockerfile.imagepack ./LargeFiles
+publish-kips:
+	@make build-kips CURRENT_DATE_TAG=${CURRENT_DATE_TAG}
```
nicholasberlin, line 148: could `publish-kips` depend on `build-kips`, like on line 162?

nicholasberlin: if we don't end up changing this, I've seen [guidance](https://stackoverflow.com/questions/53592761/understand-the-basic-concept-of-a-makefile) that `@make` should be `@${MAKE}` to ensure the same `make` is executed (perhaps it doesn't matter in our scenario)

rhysre (changes requested):
Some of these tests are actually failing (e.g. see the terminal output for `linux-image-rt-x86_64_5.16.12-1kali1.txt`). But they're all showing up as `PASS` in the summary.txt files and `run_tests.sh` is exiting with 0 in all cases. Investigating that now, but posting here just so this isn't merged before I finish taking a look.

rhysre:
Oh, pretty sure this is it:

```diff
vagrant@ubuntu-jammy:~/endpoint-dev/Kernel/Ebpf/testing$ git diff
diff --git a/testing/run_tests.sh b/testing/run_tests.sh
index 1753214..e2d1992 100755
--- a/testing/run_tests.sh
+++ b/testing/run_tests.sh
@@ -42,7 +42,7 @@ run_tests() {
     echo "BPF-check run for $# $arch kernel(s) at $(date)" > $SUMMARY_FILE
     for f in $RESULTS_DIR/*; do
         local kern=$(basename $f .txt)
-        grep "$SUMMARY_STRING" $f \
+        grep "$SUCCESS_STRING" $f \
             && echo "[PASS] $kern" >> $SUMMARY_FILE \
             || echo "[FAIL] $kern" >> $SUMMARY_FILE
     done
```

We've been grepping for the wrong string in the terminal output. `$SUMMARY_STRING` doesn't exist.

lrishi:
> Oh, pretty sure this is it:
> 
> ```diff
> vagrant@ubuntu-jammy:~/endpoint-dev/Kernel/Ebpf/testing$ git diff
> diff --git a/testing/run_tests.sh b/testing/run_tests.sh
> index 1753214..e2d1992 100755
> --- a/testing/run_tests.sh
> +++ b/testing/run_tests.sh
> @@ -42,7 +42,7 @@ run_tests() {
>      echo "BPF-check run for $# $arch kernel(s) at $(date)" > $SUMMARY_FILE
>      for f in $RESULTS_DIR/*; do
>          local kern=$(basename $f .txt)
> -        grep "$SUMMARY_STRING" $f \
> +        grep "$SUCCESS_STRING" $f \
>              && echo "[PASS] $kern" >> $SUMMARY_FILE \
>              || echo "[FAIL] $kern" >> $SUMMARY_FILE
>      done
> ```
> 
> We've been grepping for the wrong string in the terminal output. `$SUMMARY_STRING` doesn't exist.

Whoa, thats quite a miss. Any reason why this PR should be blocked for this?

Makefile
lrishi: Yup, on purpose

Makefile
lrishi: Fixing.

nicholasberlin (approved):

PR #126, opened 2022-08-22
Always set `RLIMIT_MEMLOCK` to infinity

mattnite:
Also I found the dockerignore got in the way of debugging so I removed it.


rhysre (approved):
LGTM

Can you please also remove the rlimit bump in the `main()` of `non-GPL/Events/EventsTrace/EventsTrace.c` as we no longer need it now that the lib is doing it.

lrishi (approved):

stanek-michal (approved):

PR #127, opened 2022-08-23
Add error-check for BTF load call

rhysre:
Our call to btf__load_vmlinux_btf was not being error checked with
libbpf_error(...) and returning a bad pointer on systems with no BTF,
which was then used, and caused a segfault.

Also add libEbpfEvents-level logging function that logs to stderr
if verbose libbpf logging is turned on. This will make stuff like this
much easier to catch in CI.


stanek-michal:
There's an `ebpf_log` function in Common.c already, it could clash

nicholasberlin (approved):
Assuming the potential naming collision is resolved, LGTM

rhysre:
Rebased and fixed naming issue.

stanek-michal (approved):

PR #128, opened 2022-08-23
EventProbe: use redirected_tty_write BTF info instead of tty_write

mmat11:
Use a different function to obtain signature indexes as `tty_write` BTF info are missing on ARM systems. Read more: https://github.com/elastic/ebpf/pull/116#issue-1327583872

Co-Authored-By: Rhys Rustad-Elliott <rhys.rustad-elliott@elastic.co>

### DIFF
--- a/non-GPL/Events/Lib/EbpfEvents.c
+++ b/non-GPL/Events/Lib/EbpfEvents.c
@@ -196,8 +196,8 @@ static int probe_fill_relos(struct btf *btf, struct EventProbe_bpf *obj)
 
     /* From https://github.com/elastic/ebpf/pull/116#issue-1327583872
      *
-     * tty_write BTF info is not available on ARM systems with default
-     * config.
+     * tty_write BTF info is not available on ARM64 kernels built
+     * with pahole < 1.22 due to a bug in pahole.
      * Use redirected_tty_write BTF info as function signature check
      * since it changes in the exact same version as tty_write (5.10.10-5.10.11)
      * and has the same parameters/indexes we need.
@@ -239,7 +239,9 @@ static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj
         err = err ?: bpf_program__set_autoload(obj->progs.fexit__tcp_v6_connect, false);
     }
 
-    // tty_write is not present in all supported kernels' BTF info (eg. amazonlinux2 x86_64)
+    // tty_write BTF information is not available on all supported kernels
+    // due to a pahole bug.
+    // If it is not present we can't attach a fentry/ program to it, so fallback to a kprobe.
     if (has_bpf_tramp && BTF_FUNC_EXISTS(btf, tty_write)) {
         err = err ?: bpf_program__set_autoload(obj->progs.kprobe__tty_write, false);
     } else {
--- a/non-GPL/Events/Lib/EbpfEvents.c
+++ b/non-GPL/Events/Lib/EbpfEvents.c
@@ -239,6 +239,13 @@ static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj
         err = err ?: bpf_program__set_autoload(obj->progs.fexit__tcp_v6_connect, false);
     }
 
+    // tty_write is not present in all supported kernels' BTF info (eg. amazonlinux2 x86_64)
+    if (has_bpf_tramp && BTF_FUNC_EXISTS(btf, tty_write)) {
+        err = err ?: bpf_program__set_autoload(obj->progs.kprobe__tty_write, false);
+    } else {
+        err = err ?: bpf_program__set_autoload(obj->progs.fentry__tty_write, false);
+    }
+
     // bpf trampolines are only implemented for x86. disable auto-loading of all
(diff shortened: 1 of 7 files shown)


rhysre:
Hmmm, mainline/5.19 is failing for arm64 and x86. You can ignore all failures except mainline (I'm working on those). But mainline was passing before and should keep passing.

I'm seeing the following:
```
[            ] stdout: Failed to unmarshal the following JSON: "": unexpected end of JSON input
```

Seems the testrunner is unable to unmarshal something that `EventsTrace` outputs. Digging into this further right now.

mmat11:
> Hmmm, mainline/5.19 is failing for arm64 and x86. You can ignore all failures except mainline (I'm working on those). But mainline was passing before and should keep passing.
> 
> I'm seeing the following:
> 
> ```
> [            ] stdout: Failed to unmarshal the following JSON: "": unexpected end of JSON input
> ```
> 
> Seems the testrunner is unable to unmarshal something that `EventsTrace` outputs. Digging into this further right now.

missed that, will have a look later (probably something changed in tests, I am pretty sure it used to work)

nicholasberlin (approved):

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -233,6 +239,13 @@ static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj
         err = err ?: bpf_program__set_autoload(obj->progs.fexit__tcp_v6_connect, false);
     }
 
+    // tty_write is not present in all supported kernels' BTF info (eg. amazonlinux2 x86_64)
```
rhysre, line 242: Why is this needed? It's true that `tty_write` isn't present in all supported kernel's BTF info, but that should simply be fixed by using `redirected_tty_write` as an analog. AFAICS we should never have to not load the probe.

rhysre:
@mmat11 Figured out why this is failing on mainline/5.19.

Our "5.19" kernel is actually 6.0-rc1 (check the first line logged to the serial console):

```
Booting from ROM..ESCcESC[?7lESC[2J[    0.000000] Linux version 6.0.0-rc1 (root@lrishi-el) (x86_64-linux-gnu-gcc (Debian 10.2.1-6) 10.2.1 20210110, GNU ld (GNU Binutils for Debian) 2.35.2) #1 SMP PREEMPT_DYNAMIC Tue Aug 16 12:02:02 UTC 2022
```

6.0-rc1 added this commit: https://github.com/torvalds/linux/commit/fcb14cb1bdacec5b4374fe161e83fb8208164a85

`ii->iter_type` is 6 in the invocation of the `tty_write` probe on 6.0-rc1 (i.e. `ITER_UBUF`), so we need to be reading from `ii->ubuf` instead of `ii->iov->iov_base`.

TBH though an RC kernel should not be in our testing set, and this problem is unrelated to this PR, so I'd say we can fix that in a new PR.

rhysre:
Spoke to @lrishi, we've removed the RC kernel and I've pushed a commit to use the new imagepack above.

non-GPL/Events/Lib/EbpfEvents.c
mmat11: we still want to hook `tty_write`, right? this block of code just loads a kprobe instead of the `tracing` counterpart when BTF info are missing

mmat11:
> Our "5.19" kernel is actually 6.0-rc1 (check the first line logged to the serial console):

oh.. thanks for looking into it

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -194,15 +194,21 @@ static int probe_fill_relos(struct btf *btf, struct EventProbe_bpf *obj)
...
     err = err ?: FILL_FUNC_RET_IDX(obj, btf, vfs_rename);
 
-    // TODO: Re-enable tty_write probe when BTF issues are fixed
-#if 0
-    if (FILL_FUNC_ARG_EXISTS(obj, btf, tty_write, from)) {
-        err = err ?: FILL_FUNC_ARG_IDX(obj, btf, tty_write, buf);
-        err = err ?: FILL_FUNC_ARG_IDX(obj, btf, tty_write, count);
+    /* From https://github.com/elastic/ebpf/pull/116#issue-1327583872
+     *
+     * tty_write BTF info is not available on ARM systems with default
```
rhysre, line 199: This isn't _quite_ right, it's not the default ARM64 config, it's when a (relocatable) ARM kernel is built with pahole < 1.22. Can you update this to say "tty_write BTF info is not available on ARM64 kernels built with pahole < 1.22 due to a bug in pahole".

non-GPL/Events/Lib/EbpfEvents.c
rhysre: Ok, I was a tad confused earlier. This code looks good to me. Can you clarify the comment though. I'd switch it to:

 "tty_write BTF information is not available on all supported kernels due to a pahole bug. If it is not present we can't attach a fentry/ program to it, so fallback to a kprobe"

rhysre (approved):
Two comment changes for clarity please, otherwise LGTM. 

Thanks for getting this done Mattia! 🚀

PR #130, opened 2022-08-25
Skeleton headers for host isolation

stanek-michal:
Changes required to use skeleton headers for host isolation.
It is still possible to use `*bpf.o` files and the Host Isolation demos continue to use those.

### DIFF
--- a/non-GPL/Events/EventsTrace/EventsTrace.c
+++ b/non-GPL/Events/EventsTrace/EventsTrace.c
@@ -666,7 +666,7 @@ int main(int argc, char **argv)
     if (g_features_autodetect)
         ebpf_detect_system_features(&opts.features);
 
-    err = ebpf_event_ctx__new(&ctx, event_ctx_callback, opts, 0);
+    err = ebpf_event_ctx__new(&ctx, event_ctx_callback, opts, /* dry_run= */ false);
 
     if (err < 0) {
         fprintf(stderr, "Could not create event context: %d %s\n", err, strerror(-err));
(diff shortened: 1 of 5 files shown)


non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -405,7 +405,8 @@ int ebpf_set_verbose_logging()
 
 int ebpf_event_ctx__new(struct ebpf_event_ctx **ctx,
                         ebpf_event_handler_fn cb,
-                        struct ebpf_event_ctx_opts opts)
+                        struct ebpf_event_ctx_opts opts,
+                        bool dry_run)
```
stanek-michal, line 409: This addition doesn't really change functionality but I wanted to make this option more explicit in the code to prevent bugs or relying on a certain behavior in the function. We did in fact had bugs like this in this function before. The Host Isolation version of this also has a dryRun parameter.

non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -666,7 +666,7 @@ int main(int argc, char **argv)
     if (g_features_autodetect)
         ebpf_detect_system_features(&opts.features);
 
-    err = ebpf_event_ctx__new(&ctx, event_ctx_callback, opts);
+    err = ebpf_event_ctx__new(&ctx, event_ctx_callback, opts, 0);
```
lrishi, line 669:
```suggestion
    err = ebpf_event_ctx__new(&ctx, event_ctx_callback, opts, /* dry_run= */ false);
```

Trinity2019 (approved):

rhysre (approved):

stanek-michal:
abandoning for https://github.com/elastic/ebpf/pull/233

PR #131, opened 2022-08-25
Remove kali 5.10.0 kernels from test kernels

rhysre:
Remove kali 5.10.0 kernels. The version string for this _happened_ to
contain the string "5.10.12" (as the GNU ld used to build this kernel
was built on 5.10.12) but was actually kernel 5.10.0 (which we don't
support).


lrishi (approved):
Remove asdf before merging. Thanks for the hard work on this Rhys!! 🥇

mattnite (approved):

PR #135, opened 2022-08-29
Add more mainline kernels

rhysre:
This makes the mainline kernel image test set more comprehensive. It
includes a number of kernels from the 5.10 and 5.15 LTS series instead
of just doing 5.x.0 across the board. It also adds 5.19.

The oldest kernel we can currently support is 5.10.16. On 5.10.10 (which
is the oldest we declare we can support, there is a verifier bug that
causes the probes to not load (see kernel commit
ee114dd64c0071500345439fc79dd5e0f9d106ed, went into linux 5.11 and was
added to 5.10.16 in the 5.10 LTS series). We could theoretically hack
around this bug but it's likely best value for time to just bump our
minimum declared supported version to 5.10.16.


lrishi (approved):
lgtm 👍🏽

mmat11 (approved):

PR #138, opened 2022-09-01
Run host isolation tests in GitHub Actions

rhysre:
Completely break free of Jenkins by running host isolation tests in
GitHub actions. The existing jenkins setup was broken anyways and slated
to be moved to the multikernel tester.

As a bonus of moving the host isolation tests to the new test framework,
they're now running on far more kernels than before.


testing/testrunner/utils.go
```diff
@@ -251,7 +251,13 @@ func IsOverlayFsSupported() bool {
 	return false
 }
 
-func RunTest(f func(*EventsTraceInstance), args ...string) {
+func RunTest(f func()) {
+	testFuncName := runtime.FuncForPC(reflect.ValueOf(f).Pointer()).Name()
```
mmat11, line 255: what about passing a name to `RunTest()`/`RunEventsTest()`?

Trinity2019 (approved):

mattnite (approved):
lgtm

testing/run_tests.sh
```diff
@@ -100,7 +100,7 @@ main() {
         && exit_usage
 
     local initramfs="initramfs-${arch}.cpio"
-    ./scripts/gen_initramfs.sh $arch $eventstrace $initramfs \
```
Tacklebox, line 103: If the artifacts path has a space or glob character like `[$?` etc this will have an issue. It probably should have been quoted before too. It's in a few places.
`s/$artifacts/"$artifacts"/`

testing/testrunner/utils.go
```diff
@@ -131,7 +131,7 @@ func runTestBin(binName string) []byte {
 
 	output, err := cmd.Output()
 	if err != nil {
-		fmt.Println(string(err.(*exec.ExitError).Stderr))
```
Tacklebox, line 134: Does the format of the error include stderr? Maybe we should print both?

Tacklebox (commented):
Looks good.
One nitpick that wont affect Github actions and a question.

testing/run_tests.sh
rhysre: Thanks, fixed.

testing/testrunner/utils.go
rhysre: Yeah printing the error results in  the standard error (or a subset of it if long) being printed, see here: https://pkg.go.dev/os/exec#ExitError. I've updated the printing to be a bit more verbose about this.

mattnite (approved):

lrishi (approved):
lgtm

PR #141, opened 2022-09-07
Prefer add_custom_command to add_custom_target

nicholasberlin:
Not sure why but `add_custom_target` always rebuilds non-GPL/Events/Lib/CMakeFiles/EbpfEvents.dir/EbpfEvents.c.o

Using a combo of `add_custom_command` and `add_custom_target` prevents the object file from being rebuilt when not necessary.

### DIFF
--- a/cmake/modules/BPF.cmake
+++ b/cmake/modules/BPF.cmake
@@ -78,12 +78,14 @@ function (ebpf_probe_target target)
 
     set(EBPF_PROBE_DEPFILE ${CMAKE_CURRENT_BINARY_DIR}/${target}.bpf.d)
 
-    add_custom_target(${target}_Probe
+    add_custom_command(
+        OUTPUT ${OUT_FILE} ${SKEL_FILE}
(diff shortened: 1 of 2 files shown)


rhysre (approved):

cmake/modules/BPF.cmake
```diff
@@ -78,13 +78,15 @@ function (ebpf_probe_target target)
...
-    add_custom_target(${target}_Probe
+    add_custom_command(
+        OUTPUT ${OUT_FILE} ${SKEL_FILE}
         COMMAND ${EBPF_EXTERNAL_ENV_FLAGS} ${BPF_COMPILER} ${BPF_COMPILER_FLAGS} -MD -MF ${EBPF_PROBE_DEPFILE} ${EBPF_PROBE_FLAGS} -c ${EBPF_PROBE_SOURCES} -o ${OUT_FILE}
         COMMAND ${STRIP_CMD}
         COMMAND ${SKELETON_CMD}
-        BYPRODUCTS ${OUT_FILE} ${SKEL_FILE}
     )
 
+    add_custom_target(${target}_Probe DEPENDS ${OUT_FILE} ${SKEL_FILE})
```
lrishi, line 88: This actually seems to break the dependency? How does <>_Probe know when to build OUT_FILE and SKEL_FILE?

lrishi: Might start causing issues in parallel builds :S

lrishi: oh I see, using OUTPUT. Makes sense. lgtm! thanks!

lrishi (approved):

PR #150, opened 2022-09-23
Track the probe source files

nicholasberlin:
Previously, EbpfEvents.c.o and libEbpfEvents.a would not be rebuilt if probe source files were updated, likewise for host isolation libs. This PR make cmake aware of those sources and ensures a rebuild happens on changes to those files.

### DIFF
--- a/GPL/Events/CMakeLists.txt
+++ b/GPL/Events/CMakeLists.txt
@@ -30,7 +30,7 @@ ebpf_probe_target(EventProbe
     DEPENDENCIES libbpf vmlinux
     FLAGS ${EVENTS_PROBE_CFLAGS}
     PUBLIC_HEADERS EbpfEventProto.h
-    FILES
+    DEPENDS
     ${CMAKE_CURRENT_SOURCE_DIR}/File/Probe.bpf.c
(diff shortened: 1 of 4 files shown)


GPL/Events/CMakeLists.txt
```diff
@@ -30,5 +30,15 @@ ebpf_probe_target(EventProbe
     DEPENDENCIES libbpf vmlinux
     FLAGS ${EVENTS_PROBE_CFLAGS}
     PUBLIC_HEADERS EbpfEventProto.h
+    FILES
+    ${CMAKE_CURRENT_SOURCE_DIR}/File/Probe.bpf.c
```
mmat11, line 34: I don't know how cmake works, is there a way to put a wildcard here?

mmat11 (approved):
tested locally, lgtm 👍🏻

GPL/Events/CMakeLists.txt
nicholasberlin: internet consensus seems to be explicit is better. https://stackoverflow.com/questions/1027247/is-it-better-to-specify-source-files-with-glob-or-each-file-individually-in-cmak

rhysre (approved):
👍 Thanks Nick!

PR #151, opened 2022-09-24
Move BPF support-detection logic from endpoint to this repository

rhysre:
Since the very beginning, the logic to detect if we support BPF (and all required BPF features) has been done in endpoint instead of in the ebpf events library. This is wonky. Logically, this logic belongs here, the library should be performing all necessary checks and returning `-ENOTSUP` if it can't load the probes. Library users shouldn't have to divine the specific system characteristics needed for the library to successfully run themselves.

This PR moves all the support-detection logic from endpoint to this repository, clarifies the minimum supported Linux version as 5.10.16 (not 5.10.10) and simplifies related feature detection logic. See commit messages for detailed explanations.

### DIFF
--- a/GPL/Events/Process/Probe.bpf.c
+++ b/GPL/Events/Process/Probe.bpf.c
@@ -235,8 +235,12 @@ int BPF_KPROBE(kprobe__commit_creds, struct cred *new)
     return commit_creds__enter(new);
 }
 
-static int tty_write__enter(const char *buf, ssize_t count, struct file *f)
+static int tty_write__enter(struct kiocb *iocb, struct iov_iter *from)
 {
+    const char *buf = BPF_CORE_READ(from, iov, iov_base);
+    ssize_t count   = BPF_CORE_READ(from, iov, iov_len);
+    struct file *f  = BPF_CORE_READ(iocb, ki_filp);
+
     if (is_consumer())
         goto out;
 
@@ -305,19 +309,11 @@ out:
 SEC("fentry/tty_write")
 int BPF_PROG(fentry__tty_write, struct kiocb *iocb, struct iov_iter *from)
 {
-    const char *buf = BPF_CORE_READ(from, iov, iov_base);
-    ssize_t count   = BPF_CORE_READ(from, iov, iov_len);
-    struct file *f  = BPF_CORE_READ(iocb, ki_filp);
-
-    return tty_write__enter(buf, count, f);
+    return tty_write__enter(iocb, from);
 }
 
 SEC("kprobe/tty_write")
 int BPF_KPROBE(kprobe__tty_write, struct kiocb *iocb, struct iov_iter *from)
 {
-    const char *buf = BPF_CORE_READ(from, iov, iov_base);
-    ssize_t count   = BPF_CORE_READ(from, iov, iov_len);
-    struct file *f  = BPF_CORE_READ(iocb, ki_filp);
-
-    return tty_write__enter(buf, count, f);
+    return tty_write__enter(iocb, from);
 }
--- a/GPL/Events/Process/Probe.bpf.c
+++ b/GPL/Events/Process/Probe.bpf.c
@@ -16,12 +16,6 @@
 #include "Helpers.h"
 #include "PathResolver.h"
 
-/* tty_write */
-DECL_FUNC_ARG(redirected_tty_write, iter);
-DECL_FUNC_ARG(redirected_tty_write, buf);
-DECL_FUNC_ARG(redirected_tty_write, count);
-DECL_FUNC_ARG_EXISTS(redirected_tty_write, iter);
-
 SEC("tp_btf/sched_process_fork")
 int BPF_PROG(sched_process_fork, const struct task_struct *parent, const struct task_struct *child)
 {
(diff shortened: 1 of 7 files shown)


non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -375,16 +353,79 @@ static bool system_has_bpf_tramp()
...
+
+static bool system_supports_bpf_events(void)
+{
+    bool supported    = false;
+    uint64_t features = detect_system_features();
+
+    // We only support Linux 5.10.16+
+    //
+    // Linux commit e114dd64c0071500345439fc79dd5e0f9d106ed (went in in
+    // 5.11/5.10.16) fixed a verifier buf that (as of 9/28/2022) causes our
```
nicholasberlin, line 374:
```suggestion
    // 5.11/5.10.16) fixed a verifier bug that (as of 9/28/2022) causes our
```
?

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -375,16 +353,79 @@ static bool system_has_bpf_tramp()
...
+#undef VERSION
+
+    if (!libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL)) {
+        verbose("Kernel does not support BPF_PROG_TYPE_TRACING, bpf events are not supported\n");
+        goto out;
+    }
+
+    if (!features & EBPF_FEATURE_BPF_TRAMP) {
+        // BPF trampolines are not supported, can we fallback to kprobes?
+        verbose("kernel does not support BPF trampolines, detecting kprobe support");
```
nicholasberlin, line 409:
```suggestion
        verbose("kernel does not support BPF trampolines, detecting kprobe support\n");
```

nicholasberlin (approved):
LGTM

GPL/Events/Process/Probe.bpf.c
```diff
@@ -296,62 +290,21 @@ static int tty_write__enter(const char *buf, ssize_t count, struct file *f)
...
-
-        struct kiocb *iocb = (struct kiocb *)ctx[0];
-        f                  = BPF_CORE_READ(iocb, ki_filp);
-    } else {
-        buf   = FUNC_ARG_READ(___type(buf), redirected_tty_write, buf);
-        count = FUNC_ARG_READ(___type(count), redirected_tty_write, count);
-
-        f = (struct file *)ctx[0];
-    }
+    const char *buf = BPF_CORE_READ(from, iov, iov_base);
```
mmat11, line 295: these 3 params can be read in `tty_write_enter`

non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -104,17 +96,13 @@ static const struct argp_option opts[] = {
      "Print network connection closed events", 0},
     {"print-features-on-init", 'i', NULL, false,
      "Print a message with feature information when probes have been successfully loaded", 1},
-    {"features-autodetect", 'd', NULL, false, "Autodetect features based on running kernel", 1},
```
mmat11, line 107: `testing/testrunner/eventstrace.go` uses this flag, it can be removed from there too

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -375,16 +353,79 @@ static bool system_has_bpf_tramp()
...
+    }
+
+    if (VERSION(maj, min, patch) < VERSION(5, 10, 16)) {
+        verbose("kernel version is < 5.10.16 (%d.%d.%d), bpf events are not supported\n", maj, min,
+                patch);
+        goto out;
+    }
+#undef VERSION
+
+    if (!libbpf_probe_bpf_prog_type(BPF_PROG_TYPE_TRACING, NULL)) {
```
mmat11, line 402: I don't think we should `goto out;` here, the host machine has to either support tracing programs or kprobe programs.
I'd put this feature check in the same condition of `!(features & BPF_TRAMP)` and check for kprobes in the `else` branch

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -410,13 +451,21 @@ int ebpf_set_verbose_logging()
...
+    return ctx->features;
+}
+
+int ebpf_event_ctx__new(struct ebpf_event_ctx **ctx, ebpf_event_handler_fn cb, uint64_t events)
 {
     struct EventProbe_bpf *probe = NULL;
     struct btf *btf              = NULL;
 
+    if (!system_supports_bpf_events()) {
+        verbose("this system does not support BPF (see logs)\n");
```
mmat11, line 465: nit
```suggestion
        verbose("this system does not support BPF events (see logs)\n");
```

non-GPL/Events/EventsTrace/EventsTrace.c
rhysre: 👍 done

non-GPL/Events/Lib/EbpfEvents.c
rhysre: Ok @mmat11 I've thought about this a bit more and I actually like your idea to simplify things better. 

An ironclad support check would look like the following:

```
kernel_version > 5.10.16 && (can_attach_to_fentry || can_attach_to_kprobe) && can_attach_to_btf_enabled_tracepoint && supports_btf
```

To implement this however, we'd need to do something like `system_has_bpf_tramp` for kprobe support (i.e a load *and* attach) AND for BTF enabled tracepoint support. That would increase the amount of ugly logic in here (I don't like `system_has_bpf_tramp`, but it's an unfortunate necessity).

Both of those things (loading and attaching to a BTF enabled tracepoint and kprobe) however will be checked when we actually load and attach to the probes, so IMO we can just perform some basic checks to cover most load failure cases at the start of `ebpf_event_ctx__new` and then load/attach to the probes, returning an error if the load/attach fails.

That should ensure that we're immediately exiting with a reasonable error message in cases where our probes blatantly won't work or might work but possibly mess up somehow `(kernel < 5.10.16 || !btf_is_supported)`, while still ultimately checking support for loading and attaching to kprobes/BTF tracepoints when we actually load the probes.

I've added a commit that does that. Let me know your thoughts.

mmat11: > The very first libbpf_probe_bpf_prog_type check is meant to ensure we can load our BPF enabled tracepoints

I forgot about those! :D 

> I've added a commit that does that. Let me know your thoughts.

lgtm

mmat11 (approved):

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -375,16 +363,65 @@ static bool system_has_bpf_tramp()
...
+    if (libbpf_get_error(btf)) {
+        verbose("Kernel does not support BTF, bpf events are not supported\n");
+        return false;
+    } else {
+        btf__free(btf);
+        return true;
+    }
+}
+
+static bool kernel_version_is_supported(void)
```
mmat11, line 388: I still suspect tests are failing on 5.10.x ubuntu based distros which report 5.10.0 as major.minor.patch. Could you try fetching the info from `/proc/version_signature` like libbpf does (https://github.com/libbpf/libbpf/blob/a0d1e22c7790f22809d33b3c5c1e3ded89157cc8/src/libbpf.c#L870)?

rhysre: Working on that right now. Ubuntu is an easy fix. 

Debian (and Debian-based distros like Kali) is slowing me down a bit. They report the `5.x.0` ABI version in `utsname.release` (like Ubuntu), but don't provide an upstream patch number at `/proc/version_signature`, instead they store the upstream kernel version in `utsname.version` (among other data).

rhysre: Given the differences in how to get the upstream kernel source version on different distros, I suspect it might be impossible to reliably get the upstream source for absolutely any arbitrary distro build setup, but if we hit ubuntu/debian and fallback to `uname` otherwise, I suspect that should be good enough.

rhysre: @mmat11 Fixed version detection for Ubuntu and Debian. Tests pass. Let me know if that looks good to you and I'll merge.

mmat11: Great!

PR #159, opened 2022-10-17
Implement variable-length ringbuf wire format

rhysre:
This commit introduces variable-length fields into the ringbuf wire format.

Over time, the number of fields in our wire format with a large theoretical maximum size has grown (pathnames, argv, tty output, etc.).

Every field in the wire format is of a fixed-size. This simplifies parsing of messages in userspace, but leads to huge memory waste as the maximum possible buffer size needs to be allocated, even when it's not used most of the time. For example, PATH_MAX on linux is 4096, but the pathnames we pass up are going to be a fraction of that size 99% of the time. Terminal output is an even bigger culprit for this sort of thing, a 4 KiB buffer is allocated for each tty_write event, even if the data being written to the tty is just one byte. For a high-frequency tty write event, this can easily cause us to drop events.

This commit adds a few new structures that enable us to pass up these large fields in variable-length buffers. Fields that are always of a fixed size (or variable length but so small as to be irrelevant from a memory-usage perspective, e.g. task->comm) remain as they were before, in a fixed-size part of the event. Variable length fields are packed into a flexible array member at the end of the event struct and each are preceeded by a header describing the type of field and size.

An unfortunate necessity of this approach is that we cannot use the ringbuffer reserve/commit API for events that have variable-length fields, as we won't know the size of the event in advance. Instead, we build each event with its variable-length fields in a percpu map, and then output it using bpf_ringbuf_output. This should result in a small performance hit as we're introducing an additional memory copy, but should be made up for in massive memory usage improvements. An alternative would have been to continue to use the reserve/commit API, passing up each event in fixed-size chunks, but this would just shift the extra copy to userspace, and introduce a bunch of complexity there as we'd need to have reconstruction logic in the Ebpf events library. In any case, there is no free lunch.

Logic and helpers for adding variable length fields have been added to Helpers.h and are fairly well documented. New events that need to use variable-length fields should add a ebpf_varlen_fields_start struct as their last member with a name of vl_fields. Variable-length fields can then be added to the event. For example, in the case of a file delete event:

        struct ebpf_file_delete_event *event = get_event_buffer();

        // ... Fill in fixed-length fields

        // Variable length fields
        ebpf_vl_fields__init(&event->vl_fields);
        struct ebpf_varlen_field *field;
        size_t size;

        // path
        ADD_VL_FIELD_OR_GOTO_EMIT(event, field, EBPF_VL_FIELD_PATH);
        size = ebpf_resolve_path_to_string(field->data, &p, task);
        ebpf_vl_field__set_size(&event->vl_fields, field, size);

    emit:
        bpf_ringbuf_output(&ringbuf, event, EVENT_SIZE(event), 0);

The emit label is required. The ADD_VL_FIELD_OR_GOTO_EMIT macro will jump to it if there is no more space remaining in the event buffer returned from get_event_buffer().

The use of variable-length fields also allows us to get rid of a bit of cruft. In particular:

- No more PATH_MAX * 2 sized buffers in EbpfEventProto.h that waste space and leak an implementation detail of the BPF probes to userspace
- The memset reimplementation to deal with argv has been removed as we will now know size of argv in the variable-length field header, making zeroing out a fixed-size buffer unnecessary

This commit also bumps the ringbuffer size to 4 MiB so that it's larger than the 2 MiB percpu event buffers.


GPL/Events/Helpers.h
```diff
@@ -127,32 +127,97 @@ const volatile int consumer_pid = 0;
...
-// error: A call to built-in function 'memset' is not supported.
+// This is too small for us, so instead we implement a percpu array ourselves
+// using a BPF_MAP_TYPE_ARRAY. We resize it in userspace to $(nproc) elements
+// and access it through the get_event_buffer helper, which returns the value
+// corresponding to the current processor.
+struct {
+    __uint(type, BPF_MAP_TYPE_ARRAY);
+    __uint(key_size, sizeof(int));
+    __uint(value_size, EVENT_BUFFER_SIZE);
+    __uint(max_entries, 0);
```
nicholasberlin, line 167: oh, this is zero because map is resized by user space?

GPL/Events/Helpers.h
```diff
@@ -127,32 +127,97 @@ const volatile int consumer_pid = 0;
...
+            bpf_printk("Bailing at variable-length field of type %d (event type %d), out of "      \
+                       "buffer space",                                                             \
+                       arg_type, arg_event->hdr.type);                                             \
+            goto emit;                                                                             \
+        }                                                                                          \
+                                                                                                   \
+        struct ebpf_varlen_fields_start *__vlf = &arg_event->vl_fields;                            \
+        struct ebpf_varlen_field *__field =                                                        \
+            (struct ebpf_varlen_field *)(&__vlf->data[__vlf->size & EVENT_BUFFER_SIZE_HALF_MASK]); \
+        __vlf->nfields++;                                                                          \
```
nicholasberlin, line 196: should we increment `arg_event->size`?

nicholasberlin: Seems like `ebpf_vl_field__set_size` is used for that. I wonder if they could be combined into this macro.

nicholasberlin: Oh, no it doesn't. I'm not sure what sets `arg_event->size`

stanek-michal:
> Instead, we build each event with its variable-length fields in a percpu map, and then output it using bpf_ringbuf_output. This should result in a small performance hit as we're introducing an additional memory copy

Any idea how we could measure the performance hit? On the ebpf side, can we use bpftool perf stats or something else to estimate it? I could perhaps test this end-to-end with Phoronix tests but it might be that it won't show up end-to-end.
And the performance hit would only be on the ebpf side I suppose? Or userspace as well?

GPL/Events/Helpers.h
rhysre: `arg_event->size` does not exist.  A `size` exists for the variable length fields start struct (i.e. `arg_event->vl_fields.start`) and for each `struct ebpf_varlen_field`, but `arg_event` is just an event struct (e.g. `ebpf_file_delete_event`), which does not have a `size` field.

You could combine `ebpf_vl_field__set_size` into this macro, I considered doing that, but I like the explicitness of this approach a bit better. Basically, when adding a variable length field, there are four operations to be performed. They are (in order):

1) Check if we have enough room in the buffer, if not, `goto emit`
2) Add a new variable length field and fill in the type field of the header
3) Actually fill in the variable-length field with its data (e.g. via `ebpf_resolve_path_to_string`)
4) (now that we know the size of the variable length field) update the `size` parameter of the `struct ebpf_varlen_field`

1/2 are combined into the macro, 3 is a separate function call (e.g. `ebpf_resolve_path_to_string`) and 4 is `ebpf_vl_field__set_size`. We could combine 1/2/3/4 into the same macro, but IMO having the "check for space, add field, fill in, set size" flow be super explicit like this makes the code clearer at the cost of a few more lines. I fully admit this is a matter of taste though, happy to change if there's strong disagreement.

GPL/Events/Helpers.h
rhysre: Yeah, exactly. I made it `0` so it's clear it's resized by userspace, but will add an inline comment to make it super explicit.

rhysre:
> > Instead, we build each event with its variable-length fields in a percpu map, and then output it using bpf_ringbuf_output. This should result in a small performance hit as we're introducing an additional memory copy
> 
> Any idea how we could measure the performance hit? On the ebpf side, can we use bpftool perf stats or something else to estimate it? I could perhaps test this end-to-end with Phoronix tests but it might be that it won't show up end-to-end. 

If it's visible at all in benchmarks, you should be able to see it by creating an artificially high load in terms of the new memory copy (e.g. `exec` with a 20 kb argv in a tight loop for 30 seconds), graph CPU usage over that period, and compare to the old implementation. I didn't measure that because I'd be quite surprised if it was substantial, but can get some numbers tomorrow.

Come to think of it, it would be nice to do the same thing and measure memory usage / dropped events as well to verify this PR is doing it's job (I've run a few informal tests on my laptop and it's shown a large decrease in dropped events when under load, but haven't saved the results and made them presentable).

Also purely from personal experience, I can say that the Cmd probes used variable length parameters for everything in their wire format, and I don't recall that (in BPF land, or in userspace) ever leading to a performance issue.

> And the performance hit would only be on the ebpf side I suppose? Or userspace as well?

Theoretically, there should should be a tiny perf hit in userspace from the fact that parsing the variable length fields requires a bit more math than before (see `FOR_EACH_VARLEN_FIELD`), but that's undoubtedly inconsequential (one extra memory dereference and addition per variable length parameter).

GPL/Events/EventProbe.bpf.c
```diff
@@ -15,7 +15,7 @@ char LICENSE[] SEC("license") = "Dual BSD/GPL";
 
 struct {
     __uint(type, BPF_MAP_TYPE_RINGBUF);
-    __uint(max_entries, 4096 * 64); // 256KB
+    __uint(max_entries, 1 << 26); // 4 MiB
```
stanek-michal, line 18: that's 64M not 4

GPL/Events/Helpers.h
```diff
@@ -127,32 +127,97 @@ const volatile int consumer_pid = 0;
 // From include/uapi/asm-generic/termbits.h
 #define ECHO 0x00008
 
-static bool IS_ERR_OR_NULL(const void *ptr)
-{
-    return (!ptr) || (unsigned long)ptr >= (unsigned long)-MAX_ERRNO;
-}
+// We can't use the ringbuf reserve/commit API if we want to output an event
```
mmat11, line 130: `Logic and helpers for adding variable length fields have been added to Helpers.h`

what about introducing an additional header?

non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -481,16 +536,26 @@ static void out_process_tty_write(struct ebpf_process_tty_write_event *evt)
 
     out_pid_info("pids", &evt->pids);
     out_comma();
-    out_uint("tty_out_len", evt->tty_out_len);
```
mmat11, line 484: IIRC we need len here aswell, sometimes the buffer contains more data than what should be taken

GPL/Events/Helpers.h
nicholasberlin: I like the explicitness of the separate steps. But agree consistency is most important.

GPL/Events/EventProbe.bpf.c
rhysre: Whoops, thanks for the catch.

GPL/Events/Helpers.h
rhysre: You mean adding a new variable-length field? You can do that by adding a new `ebpf_varlen_field_type` (in `EbpfEventProto.h`) and using that type in your `ADD_VL_FIELD_OR_GOTO_EMIT` invocation.

GPL/Events/Helpers.h
```diff
@@ -127,32 +127,97 @@ const volatile int consumer_pid = 0;
...
+// EVENT_BUFFER_SIZE_HALF operation below (which is necessary to keep the
+// verifier happy) will cause us to overwrite preceeding variable length
+// fields, resulting in corrupt data
+#define ADD_VL_FIELD_OR_GOTO_EMIT(arg_event, arg_field, arg_type)                                  \
+    do {                                                                                           \
+        if (EVENT_SIZE(arg_event) >= EVENT_BUFFER_SIZE_HALF) {                                     \
+            bpf_printk("Bailing at variable-length field of type %d (event type %d), out of "      \
+                       "buffer space",                                                             \
+                       arg_type, arg_event->hdr.type);                                             \
+            goto emit;                                                                             \
```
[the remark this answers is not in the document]
rhysre: Yes, this is the case. We would never encounter this "out of space in the buffer" case previously, as we were guaranteed to have space available by virtue of bpf_ringbuf_reserve. In this new implementation, we can either not emit the event, or emit it with partial data if we don't have enough space in the buffer.

Given the lowish limits put on the largest possible variable length fields (tty out, argv), and the large size of the percpu event buffers, we should currently never hit this branch, but could in the future if more large fields are added. IMO it would generally make sense to emit what we have in a best effort fashion in that case, which is why I wrote this in this way.

non-GPL/Events/EventsTrace/EventsTrace.c
rhysre: Length will still be passed up. Since `tty_out` is now a variable length field, we have the length of tty_out in the `struct ebpf_varlen_field` struct preceding it (`size` attribute).

GPL/Events/Helpers.h
mmat11: no, I meant having all "vl field" related logic in a dedicated header file instead of using Helpers.h

non-GPL/Events/EventsTrace/EventsTrace.c
mmat11: oh true, I forgot we are reading up to `len`

rhysre:
Ok some quick and dirty performance numbers here.

In my Vagrant dev VM, I fired up EventsTrace as follows:
```
 sudo ./artifacts-x86_64/non-GPL/Events/EventsTrace/EventsTrace -a > /dev/null
```

Then fired up a loop spam `exec`'ing `ls` with a very long `argv`: 
```
 while true; do ls $(python3 -c 'print("a" * 10000)'); done
```

This results in both a huge number of exec's with long `argv`'s and a ton of data written to the terminal, so lots of exec and tty_write events with large variable length fields. Should be about as stressful as you can get for a single-core test w.r.t. the newly added memory copy from `bpf_ringbuf_output`.

CPU time (both in-kernel and in-userspace) and memory usage were measured when the spam loop had already started for 60s with this script:

```bash
echo 'time (s), %user, %sys, usedmem'

time=0
while true; do
    cpu=$(top -n 1 -b | head -3 | tail -1 | awk '{print $2 "," $4}')
    usedmem=$(free | head -2 | tail -1 | awk '{print $3}')
    echo $time,$cpu,$usedmem
    sleep 1
    time=$(echo $time+1 | bc)

    if [[ $time == "60" ]]; then
        break;
    fi
done
```

From a quick glance, it looks like time spent executing in BPF programs in kernelspace isn't attributed to the process that loaded/attached those programs, hence the system wide measurement with `top` instead of a per-process value. Should add some noise, but we will be picking up the extra work done by the BPF program that way.

Graphs of results are as follows (old = fixed length, new = variable length)

Variable-Length Fields vs  Fixed-Length_ CPU Usage

Variable-Length Fields vs  Fixed-Length_ Memory Usage

CPU usage doesn't seem to be measurably different in kernelspace, which is where the extra copy is. Hard to tell, but there might be a tiny bump in userspace CPU time from the extra parsing. In any case, the effect if it exists is very small.

Memory usage is ~15MB higher, which makes sense. My VM has 5 cores, and there's an extra 2MB of locked memory added per core from the per-cpu map, plus another ~4MB from the ringbuffer size bump, so about 15 MB total. 

Given the caps I put on argv and tty_out (20 kb and 8 kb respectively), 2MB for each percpu buffer is excessive as it's impossible to hit that for our largest feasible event. Originally, I chose an arbitrary, fairly big number just so we didn't have to worry about overrunning it in the future, but will drop it to 256 KiB as that is still more than enough to accommodate our largest-sized event and will and reduce memory usage on systems with a tons of cores.

GPL/Events/Helpers.h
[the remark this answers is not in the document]
mmat11: > IMO it would generally make sense to emit what we have in a best effort fashion in that case

not sure about this, it depends on the data that is being (not) sent. For example, it wouldn't make sense to emit a tty_write event without buffer while it could make sense to emit an exec event without cgroup path.
Also, our user space code currently assumes all this data is present

GPL/Events/Helpers.h
rhysre: Oh good idea actually, they don't really fit in Helpers.h. Moved to Varlen.h.

GPL/Events/Helpers.h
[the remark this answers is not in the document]
rhysre: Ok I've thought this over and realized the whole discussion is basically a moot point.

The largest possible event we could send up (an exec with a 20 KiB argv and `PATH_MAX` sized filename, cwd and pids ss cgroup path), is no more than 50 KiB in size. We have 128 KiB of useable space in the percpu event buffers, meaning currently, the `goto emit` path will never be hit.

When I originally wrote this, the `goto emit` path could have been hit, because I was putting huge limits on the max size of the `argv` and `tty_out` that we read (~1/2 the size of the percpu event buffer, the maximum size the verifier would accept). I later reduced this to 20 KiB for `argv` and 8 KiB for `tty_out` because it would have been too easy to DoS endpoint otherwise by spamming an exec with a huge argv or performing a ton of tty writes. The kprobe event implementation also caps argv at a modest value IIRC, so there was precedent for that.

As a side effect of the limits on the size of argv and tty_out, as mentioned, the max event size is now ~40-50 KiB, meaning the `goto emit` path will never be hit. I didn't realize that when I put those limits in, so the `goto emit` logic remained. Now however, I can see that, for the time being, we simply don't need that logic at all as we will never bail out and emit an event with missing fields.

If there's a need for reading truly massive variable length fields in the future, we can revisit this and perhaps add this logic back, but it's been removed for now.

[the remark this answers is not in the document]
rhysre: Thanks by the way @mmat11 for pointing this out and making me come to a realization that simplified the code 🙏

stanek-michal:
> Ok some quick and dirty performance numbers here.
> 

Awesome, love the charts! Interesting that it seems to be higher on the user CPU time, but it's close enough that it could just be within error.

> Given the caps I put on argv and tty_out (20 kb and 8 kb respectively), 2MB for each percpu buffer is excessive as it's impossible to hit that for our largest feasible event. Originally, I chose an arbitrary, fairly big number just so we didn't have to worry about overrunning it in the future, but will drop it to 256 KiB as that is still more than enough to accommodate our largest-sized event and will and reduce memory usage on systems with a tons of cores.

Cool. After all, EC2 has instances with over 440 cores (although those would almost certainly have much more RAM than 1GB)

GPL/Events/Helpers.h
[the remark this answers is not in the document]
mmat11: nice!

mmat11 (approved):
lgtm,nice work!

GPL/Events/File/Probe.bpf.c
```diff
@@ -370,12 +388,25 @@ static int vfs_rename__exit(int ret)
...
 
-    bpf_ringbuf_submit(event, 0);
+    // Variable length fields
+    ebpf_vl_fields__init(&event->vl_fields);
+    struct ebpf_varlen_field *field;
+    size_t size;
+
+    // old path
+    field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_OLD_PATH);
+    size  = bpf_probe_read_kernel_str(field->data, PATH_MAX, ss->rename.old_path);
```
stanek-michal, line 401: I know we weren't doing it before but I feel like we really should be checking the return value from `bpf_probe_read_kernel_str`. It can return a negative value which would overflow here. To avoid complicating the code too much I suppose we should just `__set_size()` to 0 if we detect error here.

GPL/Events/File/Probe.bpf.c
```diff
@@ -370,12 +388,25 @@ static int vfs_rename__exit(int ret)
...
+    ebpf_vl_field__set_size(&event->vl_fields, field, size);
+
+    // new path
+    field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_NEW_PATH);
+    size  = bpf_probe_read_kernel_str(field->data, PATH_MAX, ss->rename.new_path);
```
stanek-michal, line 406: same comment as above

GPL/Events/Process/Probe.bpf.c
```diff
@@ -70,12 +89,33 @@ int BPF_PROG(sched_process_exec,
...
+    ebpf_vl_field__set_size(&event->vl_fields, field, size);
+
+    // cwd
+    field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_CWD);
+    size  = ebpf_resolve_path_to_string(field->data, &task->fs->pwd, task);
+    ebpf_vl_field__set_size(&event->vl_fields, field, size);
+
+    // filename
+    field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_FILENAME);
+    size  = bpf_probe_read_kernel_str(field->data, PATH_MAX, binprm->filename);
```
stanek-michal, line 115: needs error check

stanek-michal (approved):
Nice work, LGTM!
I would only add error checks after bpf_probe_read* calls to be safe.

GPL/Events/File/Probe.bpf.c
stanek-michal: I also wonder if this would leak 128K of kernel data in case of such error, looks like it might

rhysre: Ty for the catch. Fixed, all kernel string reads are now error checked and an empty string is passed up on error.

GPL/Events/Process/Probe.bpf.c
rhysre: fixed

PR #162, opened 2022-11-21
Fetch env from struct mm for exec events

lrishi:
Elastic Defend supports reporting envvars for newly spawned processes.
Currently, all envvars are gathered in userspace, which is a best effort mechanism, and prone to race conditions.
For example: a process can exit before the userspace application can get to fetching all envvars.

To prevent this and make the feature more resilient, we can gather envvars from struct mm* and add it to the exec event metadata.


non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -306,6 +306,28 @@ static void out_argv(const char *name, char *buf, size_t buf_size)
     out_string(name, scratch_space);
 }
 
+static void out_env(const char *name, char *buf, size_t buf_size)
+{
+    // Buf is the argv array, with each argument delimited by a '\0', rework
```
nicholasberlin, line 311:
```suggestion
    // Buf is the env array, with each argument delimited by a '\0', rework
```

nicholasberlin: Could probably rename and reuse `out_argv`, as they seem identical to me.

nicholasberlin (approved):

testing/test_bins/fork_exec.c
```diff
@@ -27,7 +27,8 @@ int main()
         gen_pid_info_json(pid_info, sizeof(pid_info));
         printf("{ \"parent_info\": %s, \"child_pid\": %d}\n", pid_info, pid);
     } else {
-        CHECK(execl("./do_nothing", "./do_nothing", NULL), -1);
+        char *env_list[] = {"TEST_ENV_KEY1=TEST_ENV_VAL1", "TEST_ENV_KEY2=TEST_ENV_VAL2", NULL};
```
rhysre, line 30: +1 Thank you for adding tests

rhysre (approved):

PR #166, opened 2023-01-31
Add support for Trusted Applications

stanek-michal:
Do not create file and network events for trusted processes

### DIFF
--- a/GPL/Events/File/Probe.bpf.c
+++ b/GPL/Events/File/Probe.bpf.c
@@ -36,6 +36,9 @@ static int do_unlinkat__enter()
 {
     struct ebpf_events_state state = {};
     state.unlink.step              = UNLINK_STATE_INIT;
+    if (ebpf_events_is_trusted_pid(0)) {
+        return 0;
+    }
     ebpf_events_state__set(EBPF_EVENTS_STATE_UNLINK, &state);
     return 0;
 }
@@ -56,6 +59,9 @@ static int mnt_want_write__enter(struct vfsmount *mnt)
 {
     struct ebpf_events_state *state = NULL;
 
+    if (ebpf_events_is_trusted_pid(0)) {
+        goto out;
+    }
     state = ebpf_events_state__get(EBPF_EVENTS_STATE_UNLINK);
     if (state) {
         // Certain filesystems (eg. overlayfs) call mnt_want_write
@@ -99,6 +105,8 @@ int BPF_KPROBE(kprobe__mnt_want_write, struct vfsmount *mnt)
 static int vfs_unlink__exit(int ret)
 {
     if (ret != 0)
+        goto out;
+    if (ebpf_events_is_trusted_pid(0))
         goto out;
 
     struct ebpf_events_state *state = ebpf_events_state__get(EBPF_EVENTS_STATE_UNLINK);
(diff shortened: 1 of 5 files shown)


non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -683,3 +683,97 @@ void ebpf_event_ctx__destroy(struct ebpf_event_ctx **ctx)
...
+    {
+        verbose("Ebpf trusted pids map is NULL");
+        return NULL;
+    }
+    return map;
+}
+
+static int ebpf_clear_process_trustlist(int map_fd)
+{
+    int rv                   = 0;
```
nicholasberlin, line 711: this variable doesn't seem necessary.

nicholasberlin (approved):

nicholasberlin:
CI is saying there are code format issues.

Trinity2019 (approved):

GPL/Events/File/Probe.bpf.c
```diff
@@ -56,6 +59,9 @@ static int mnt_want_write__enter(struct vfsmount *mnt)
 {
     struct ebpf_events_state *state = NULL;
 
+    if (ebpf_events_is_trusted_pid(0)) {
```
mmat11, line 62: I don't think this is needed here as this hook point is part of multiple chains but never the first one to execute

mmat11: In general, for file probes, IIRC the entrypoints are: `do_unlinkat__enter`, `do_filp_open__exit` and `do_renameat2__enter` so 3 checks should be sufficient

GPL/Events/Network/Probe.bpf.c
```diff
@@ -22,6 +22,8 @@ static int inet_csk_accept__exit(struct sock *sk)
 {
```
mmat11, line 22: for network probes, the entrypoints should be: `inet_csk_accept__exit`, `tcp_v*_connect` (but you can use `tcp_connect`) and `tcp_close__enter`

GPL/Events/File/Probe.bpf.c
stanek-michal: You're right, doesn't make sense to check trusted pids before `state__get()` call because the function is going to return from that anyway. I removed the redundant calls, please check if it looks right.

GPL/Events/State.h
```diff
@@ -137,4 +137,25 @@ static long ebpf_events_scratch_space__set(enum ebpf_events_state_op op,
...
+/* Trusted Apps - list of trusted pids */
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(key_size, sizeof(u32));
+    __uint(value_size, sizeof(u32));
+    __uint(max_entries, 512);
+} elastic_ebpf_events_trusted_pids SEC(".maps");
+
+/* Trusted Apps - check if current pid is trusted for given event type */
+static bool ebpf_events_is_trusted_pid(int type)
```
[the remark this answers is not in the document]
stanek-michal: removed it

mmat11 (approved):

GPL/Events/File/Probe.bpf.c
mmat11: lgtm!

PR #167, opened 2023-02-10
Align with libbpf function names

nicholasberlin:
This PR exposes a *__consume function
And renames *__next to *__poll

The main idea here is to allow users to avoid a second polling call, if polling is done outside this library.

### DIFF
--- a/non-GPL/Events/Lib/EbpfEvents.c
+++ b/non-GPL/Events/Lib/EbpfEvents.c
@@ -654,12 +654,21 @@ out_destroy_probe:
     return err;
 }
 
-int ebpf_event_ctx__next(struct ebpf_event_ctx *ctx, int timeout)
+int ebpf_event_ctx__poll(struct ebpf_event_ctx *ctx, int timeout)
 {
     if (!ctx)
         return -1;
 
     int consumed = ring_buffer__poll(ctx->ringbuf, timeout);
+    return consumed > 0 ? 0 : consumed;
+}
+
+int ebpf_event_ctx__consume(struct ebpf_event_ctx *ctx)
+{
+    if (!ctx)
+        return -1;
+
+    int consumed = ring_buffer__consume(ctx->ringbuf);
     return consumed > 0 ? 0 : consumed;
 }
 
--- a/non-GPL/Events/Lib/EbpfEvents.c
+++ b/non-GPL/Events/Lib/EbpfEvents.c
@@ -654,7 +654,7 @@ out_destroy_probe:
     return err;
 }
 
-int ebpf_event_ctx__poll(struct ebpf_event_ctx *ctx, int timeout)
+int ebpf_event_ctx__next(struct ebpf_event_ctx *ctx, int timeout)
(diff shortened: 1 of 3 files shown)


non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -663,6 +663,15 @@ int ebpf_event_ctx__next(struct ebpf_event_ctx *ctx, int timeout)
     return consumed > 0 ? 0 : consumed;
 }
 
+int ebpf_event_ctx__consume(struct ebpf_event_ctx *ctx)
+{
+    if (!ctx)
+        return -1;
+
+    int consumed = ring_buffer__consume(ctx->ringbuf);
+    return consumed > 0 ? 0 : consumed;
```
stanek-michal, line 672: Maybe just return `consumed` directly? I don't see what's the benefit of removing this information here, especially if it's supposed to be a generic library.

nicholasberlin: First, I didn't put much thought into it, I just copied what `ebpf_event_ctx__next` (now `ebpf_event_ctx__poll`) had done. Lazy on my part.

In my initial local attempts, I had thought to use `timeout` as a switch between `ring_buffer__consume` and `ring_buffer__poll`, and make adjustments to `ebpf_event_ctx__next`. Something like this:
```
    int consumed = 0;
    if (!ctx)
        return -1;

    if (timeout == 0)
    {
        consumed = ring_buffer__consume(ctx->ringbuf);
    }
    else
    {
        consumed = ring_buffer__poll(ctx->ringbuf, timeout);
    }
    return consumed > 0 ? 0 : consumed;
```
The trouble there is that `ring_buffer__consume` attempts to read from all potential ring buffers, where calling into `ring_buffer__poll` (and therefore `epoll_wait`) only acts up ring buffers which are ready.

If there's only one ring buffer, this scenario would indeed save some cycles, but multiple ring buffers would lead to memory barrier actions on ring buffers which haven't signaled being ready. I'm not sure whether that is preferable to a system call.

So, being lazy again, I punted and decided to expose a new function for users that know they have one ring buffer and are polling externally.

How about this ...
three functions:
1) `ebpf_event_ctx__consume` which is a pass through to `ring_buffer__consume` as you suggested.
2) `ebpf_event_ctx__poll` which is a pass through to `ring_buffer__poll` and returns `consumed` similar to 1.
3) `ebpf_event_ctx__next` which clamps the return value to 0 for success, and returns negative results from libbpf. Essentially, don't change it at all.

Adding 1 and 2 leaks the underlying implementation a bit. Not great, but potentially more efficient. 

What do you think?

nicholasberlin: sounds good, https://github.com/elastic/ebpf/pull/167/commits/d9345c8dafe8aaadb508573b7798c918ab4426d6

stanek-michal (approved):

mmat11 (approved):

PR #168, opened 2023-03-09
Update libbpf to support 6.0+ kernels.

nicholasberlin:
Starting in 6.0 BTF added a new kind: https://elixir.bootlin.com/linux/v6.0/source/tools/lib/bpf/btf.h#L319
Our current BTF handling fails when it encounters these type of entries, and thus our probes fail to load.
This PR fixes this by updating our libbpf to v1.1.0.

There are three initial commits:

6b2668b843880faa1d1f3e86839d86cc5e8a29d3
----
 This commit is essentially `tar -xf v.1.1.0.tar.gz`, and unfortunately, is a rubber stamp.

f80252032c63b4d0993f4439d059c9bb2684a6c5
----
adds our older elftoolchain headers and comments out usdt (Userland Statically Defined Tracing) handling from libbpf. elftoolchain doesn't support `gelf_getnote` which is needed by usdt handling. Since we don't leverage usdt, it seemed reasonable to just comment out the usage. Also, this commit upgrades some deprecated function use in unit tests.

f588fd78820e0cd4c6da4455e78964e978669647
----
When testing the above commits it was revealed that our tty_write probes were failing. Previously, when a buffer was passed to write and an iovec needed to be generated in kernel, `nr_segs` of iov_iter would be set to 1. Starting in 6.0, a new field was added to the `iov_iter`, [`ubuf`](https://elixir.bootlin.com/linux/v6.0/source/include/linux/uio.h#L54). Here's the [commit](https://github.com/torvalds/linux/commit/fcb14cb1bdacec5b4374fe161e83fb8208164a85). Later this [commit](https://github.com/torvalds/linux/commit/3e20a751aff0e099cff496511fef8cdf655b3360), stopped initializing `nr_segs` to one. So, single segmented iovecs were being ignored by our tty_write probe.


rhysre (approved):
Needs a `make format`, but other than that looks good to me.

mmat11 (approved):
lgtm!

stanek-michal (approved):
ACK, just make sure libbpf builds with -fPIC

PR #169, opened 2023-04-19
Compile libz with -O3

stanek-michal:
The side effect of us overwriting libz CFLAGS to add -fPIE was removal of default -O3 optimization level making libz effectively build with -O0. Restore -O3.


nfritts (approved):

nicholasberlin (approved):

stanek-michal:
Looks like no new failures, merging

PR #170, opened 2023-04-20
Try not installing epel packages

stanek-michal:
aarch64 CI build times out likely because of EPEL


nicholasberlin (approved):
LGTM if it works.

PR #172, opened 2023-08-25
Add effective and permitted caps to fork and exec events.

nicholasberlin:


testing/testrunner/tests.go
```diff
@@ -130,6 +130,10 @@ func TestForkExec(et *EventsTraceInstance) {
 		}
 	}
 
+	AssertUint64Equal(uint64(forkEvent.Creds.CapEffective), uint64(0x00000000f0f0f0f0))
```
mjwolf, line 133: This should be `CapPermitted`? you have the same line twice now

nicholasberlin: Actually fixed here: https://github.com/elastic/ebpf/pull/172/commits/66bd30a02feb8ace0f6f52290bf36f79f8a03106

mjwolf (approved):
LGTM!

GPL/Events/Helpers.h
```diff
@@ -235,6 +235,37 @@ static void ebpf_cred_info__fill(struct ebpf_cred_info *ci, const struct task_st
     ci->rgid = BPF_CORE_READ(task, cred, gid.val);
     ci->egid = BPF_CORE_READ(task, cred, egid.val);
     ci->sgid = BPF_CORE_READ(task, cred, sgid.val);
+
+    if (bpf_core_field_exists(task->cred->cap_permitted.cap)) {
```
stanek-michal, line 239: I understand this assumes that if permitted field exists, then effective also exists.

nicholasberlin: This check is to determine when the `kernel_cap_t` definition changed in https://github.com/torvalds/linux/commit/f122a08b197d076ccf136c73fae0146875812a88

Previously it was:
```
typedef struct kernel_cap_struct {
        __u32 cap[_KERNEL_CAPABILITY_U32S];
} kernel_cap_t;
```

Currently it is:
```
typedef struct { u64 val; } kernel_cap_t;
```

norrietaylor: This is a great nugget of information that isn't obvious from the code. Perhaps a comment explaining the check would be useful to our future-selves?

nicholasberlin: 40d34645171e21110675f684642f13e46d0c5c22

stanek-michal (approved):

nicholasberlin:
Thanks for the reviews!

PR #178, opened 2023-10-24
EventProbe: capture file info from inode

mmat11:
tested manually (eventstrace):

```
{"event_type":"FILE_CREATE","pids":{"tid":3874,"tgid":3813,"ppid":2956,"pgid":2956,"sid":2956,"start_time_ns":50065645985},"mount_namespace":4026531841,"comm":"Cache2 I/O","file_info":{"type":"FILE","inode":48667076,"mode":100600,"size":0,"uid":1000,"gid":1000,"mtime":493949622,"ctime":493949622},"path":"/home/matt/.cache/mozilla/firefox/lpqgi4lp.default-release/cache2/entries/080AE6076F29C7973BFF7A893740046655644EBE","symlink_target_path":""}
{"event_type":"FILE_DELETE","pids":{"tid":777527,"tgid":777527,"ppid":13733,"pgid":777527,"sid":777527,"start_time_ns":181390102798979},"mount_namespace":4026531841,"comm":"zsh","file_info":{"type":"SYMLINK","inode":48667077,"mode":120777,"size":23,"uid":1000,"gid":1000,"mtime":773993000,"ctime":774993006},"path":"/home/matt/.zsh_history.LOCK","symlink_target_path":""}
{"event_type":"FILE_DELETE","pids":{"tid":979918,"tgid":979918,"ppid":777527,"pgid":979918,"sid":777527,"start_time_ns":231767498246356},"mount_namespace":4026531841,"comm":"rm","file_info":{"type":"SYMLINK","inode":6706,"mode":120777,"size":11,"uid":1000,"gid":1000,"mtime":241817035,"ctime":776993018},"path":"/tmp/ciao124","symlink_target_path":"/tmp/ciao123"}
```

draft: this temporarily reintroduces this bug: https://github.com/elastic/ebpf/commit/039ceef2f91a1c667319a211830692b487c1cda2 ; can be merged after I find a workaround

### DIFF
--- a/GPL/Events/File/File.h
+++ b/GPL/Events/File/File.h
@@ -1,0 +1,79 @@
+// SPDX-License-Identifier: GPL-2.0-only OR BSD-2-Clause
+
+/*
+ * Copyright (C) 2021 Elasticsearch BV
+ *
+ * This software is dual-licensed under the BSD 2-Clause and GPL v2 licenses.
+ * You may choose either one of them if you use this software.
+ */
+
+#ifndef EBPF_EVENTPROBE_FILE_H
+#define EBPF_EVENTPROBE_FILE_H
+
+#include "EbpfEventProto.h"
+
+#define PATH_MAX 4096
+
+// include/uapi/linux/stat.h
+#define S_IFMT 00170000
+#define S_IFSOCK 0140000
+#define S_IFLNK 0120000
+#define S_IFREG 0100000
+#define S_IFBLK 0060000
+#define S_IFDIR 0040000
+#define S_IFCHR 0020000
+#define S_IFIFO 0010000
+#define S_ISUID 0004000
+#define S_ISGID 0002000
+#define S_ISVTX 0001000
+
+#define S_ISLNK(m) (((m)&S_IFMT) == S_IFLNK)
+#define S_ISREG(m) (((m)&S_IFMT) == S_IFREG)
+#define S_ISDIR(m) (((m)&S_IFMT) == S_IFDIR)
+#define S_ISCHR(m) (((m)&S_IFMT) == S_IFCHR)
+#define S_ISBLK(m) (((m)&S_IFMT) == S_IFBLK)
+#define S_ISFIFO(m) (((m)&S_IFMT) == S_IFIFO)
+#define S_ISSOCK(m) (((m)&S_IFMT) == S_IFSOCK)
+
+#define S_IRWXU 00700
+#define S_IRUSR 00400
+#define S_IWUSR 00200
+#define S_IXUSR 00100
+
+#define S_IRWXG 00070
+#define S_IRGRP 00040
+#define S_IWGRP 00020
+#define S_IXGRP 00010
+
+#define S_IRWXO 00007
+#define S_IROTH 00004
+#define S_IWOTH 00002
+#define S_IXOTH 00001
+
+static int ebpf_file_info__fill(struct ebpf_file_info *finfo, struct dentry *de)
+{
+    struct inode *ino = BPF_CORE_READ(de, d_inode);
+
+    finfo->inode = BPF_CORE_READ(ino, i_ino);
+    finfo->mode  = BPF_CORE_READ(ino, i_mode);
+    finfo->size  = BPF_CORE_READ(ino, i_size);
+    finfo->uid   = BPF_CORE_READ(ino, i_uid.val);
+    finfo->gid   = BPF_CORE_READ(ino, i_gid.val);
+    finfo->mtime = BPF_CORE_READ(ino, i_mtime.tv_nsec);
+    finfo->ctime = BPF_CORE_READ(ino, i_ctime.tv_nsec);
+
+    if (S_ISREG(finfo->mode)) {
+        finfo->type = EBPF_FILE_TYPE_FILE;
+    } else if (S_ISDIR(finfo->mode)) {
+        finfo->type = EBPF_FILE_TYPE_DIR;
+    } else if (S_ISLNK(finfo->mode)) {
+        finfo->type = EBPF_FILE_TYPE_SYMLINK;
+    } else {
+        bpf_printk("unknown file type (mode=%d)", finfo->mode);
+        return -1;
+    }
+
+    return 0;
+}
+
+#endif // EBPF_EVENTPROBE_FILE_H
--- a/GPL/Events/File/Probe.bpf.c
+++ b/GPL/Events/File/Probe.bpf.c
@@ -13,6 +13,7 @@
 #include <bpf/bpf_helpers.h>
 #include <bpf/bpf_tracing.h>
 
+#include "File.h"
 #include "Helpers.h"
 #include "PathResolver.h"
 #include "State.h"
@@ -123,10 +124,15 @@ static int vfs_unlink__exit(int ret)
     ebpf_pid_info__fill(&event->pids, task);
 
     struct path p;
-    p.dentry     = state->unlink.de;
+    p.dentry     = &state->unlink.de;
     p.mnt        = state->unlink.mnt;
     event->mntns = mntns(task);
     bpf_get_current_comm(event->comm, TASK_COMM_LEN);
+
+    if (ebpf_file_info__fill(&event->finfo, p.dentry)) {
+        bpf_printk("vfs_unlink__exit: failed to fill file info\n");
+        goto out;
+    }
 
     // Variable length fields
     ebpf_vl_fields__init(&event->vl_fields);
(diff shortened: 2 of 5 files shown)


GPL/Events/File/File.h
```diff
@@ -0,0 +1,79 @@
...
+static int ebpf_file_info__fill(struct ebpf_file_info *finfo, struct dentry *de)
+{
+    struct inode *ino = BPF_CORE_READ(de, d_inode);
+
+    finfo->inode = BPF_CORE_READ(ino, i_ino);
+    finfo->mode  = BPF_CORE_READ(ino, i_mode);
+    finfo->size  = BPF_CORE_READ(ino, i_size);
+    finfo->uid   = BPF_CORE_READ(ino, i_uid.val);
+    finfo->gid   = BPF_CORE_READ(ino, i_gid.val);
+    finfo->mtime = BPF_CORE_READ(ino, i_mtime.tv_nsec);
```
nicholasberlin, line 62: I'm fairly sure this only grabs the sub-second portion of the modified time.
If the file were last modified at `1001.0045 seconds`, only the `.00045` would be captured.
Assuming I'm correct, we should have two time fields for each time type in `ebpf_file_info`.
Like, `mtime_sec` and `mtime_nsec` or something.

```
finfo->mtime_sec = BPF_CORE_READ(ino, i_mtime.tv_sec);
finfo->mtime_nsec = BPF_CORE_READ(ino, i_mtime.tv_nsec);
```


Same for `i_ctime`.

Could grab `i_atime` as well, right?

GPL/Events/File/File.h
```diff
@@ -0,0 +1,79 @@
...
+    finfo->ctime = BPF_CORE_READ(ino, i_ctime.tv_nsec);
+
+    if (S_ISREG(finfo->mode)) {
+        finfo->type = EBPF_FILE_TYPE_FILE;
```
nicholasberlin, line 66: Is this struct member (`finfo->type`) necessary?

GPL/Events/File/File.h
```diff
@@ -0,0 +1,79 @@
...
+    } else if (S_ISDIR(finfo->mode)) {
+        finfo->type = EBPF_FILE_TYPE_DIR;
+    } else if (S_ISLNK(finfo->mode)) {
+        finfo->type = EBPF_FILE_TYPE_SYMLINK;
+    } else {
+        bpf_printk("unknown file type (mode=%d)", finfo->mode);
```
nicholasberlin, line 72: How noisy is this? Would hit for every socket and character device, right?
I can never remember ... is this the logging function that is disabled based on build type?

GPL/Events/File/File.h
```diff
@@ -0,0 +1,79 @@
...
+#define S_IXGRP 00010
+
+#define S_IRWXO 00007
+#define S_IROTH 00004
+#define S_IWOTH 00002
+#define S_IXOTH 00001
+
+static int ebpf_file_info__fill(struct ebpf_file_info *finfo, struct dentry *de)
+{
+    struct inode *ino = BPF_CORE_READ(de, d_inode);
```
nicholasberlin, line 55: Does `BPF_CORE_READ` protect against NULL derefs?
What even happens if an ebpf prog derefs NULL?

GPL/Events/File/Probe.bpf.c
```diff
@@ -123,11 +124,16 @@ static int vfs_unlink__exit(int ret)
     ebpf_pid_info__fill(&event->pids, task);
 
     struct path p;
-    p.dentry     = state->unlink.de;
+    p.dentry     = &state->unlink.de;
     p.mnt        = state->unlink.mnt;
     event->mntns = mntns(task);
     bpf_get_current_comm(event->comm, TASK_COMM_LEN);
 
+    if (ebpf_file_info__fill(&event->finfo, p.dentry)) {
```
nicholasberlin, line 132: I notice that we don't check the return of `ebpf_pid_info__fill` ...

As is, `ebpf_file_info__fill` will return non-zero if the dentry is not a dir, file, or symlink. Are we intentionally bailing in those cases? If so, why? Add a comment perhaps.

GPL/Events/File/Probe.bpf.c
```diff
@@ -138,6 +144,12 @@ static int vfs_unlink__exit(int ret)
     size  = ebpf_resolve_path_to_string(field->data, &p, task);
     ebpf_vl_field__set_size(&event->vl_fields, field, size);
 
+    // symlink_target_path
+    field      = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_SYMLINK_TARGET_PATH);
+    char *link = BPF_CORE_READ(p.dentry, d_inode, i_link);
+    size       = read_kernel_str_or_empty_str(field->data, PATH_MAX, link);
```
nicholasberlin, line 150: I think I understand ... 

`EBPF_VL_FIELD_SYMLINK_TARGET_PATH` field will be an empty string for non-symlinks. And, in the case of symlinks, `EBPF_VL_FIELD_PATH` is the path to the symlink being unlinked.

right?

nicholasberlin (commented):
Still going through this ...

GPL/Events/File/File.h
mmat11: I think you are correct

> Assuming I'm correct, we should have two time fields for each time type in ebpf_file_info.
Like, mtime_sec and mtime_nsec or something.

what about storing it as nanoseconds by converting the `_sec` portion to nanoseconds and adding it to the `_nsec` portion?

> Could grab i_atime as well, right?

I am only grabbing data indicated here https://github.com/elastic/integrations/issues/7401#issuecomment-1757977436

GPL/Events/File/File.h
mmat11: not 100% necessary but it comes handy

GPL/Events/File/File.h
mmat11: > is this the logging function that is disabled based on build type?

yes

GPL/Events/File/Probe.bpf.c
mmat11: > As is, ebpf_file_info__fill will return non-zero if the dentry is not a dir, file, or symlink. Are we intentionally bailing in those cases? If so, why?

I think we should preserve the original functionality (so support all file types) and add the missing types other than file, dir, symlink. It this function wasn't failing earlier, it shouldn't fail now

GPL/Events/File/Probe.bpf.c
mmat11: correct

mmat11: there's an example in the original comment:
```
...{"type":"SYMLINK","inode":6706,"mode":120777,"size":11,"uid":1000,"gid":1000,"mtime":241817035,"ctime":776993018},"path":"/tmp/ciao124","symlink_target_path":"/tmp/ciao123"}
```
which was triggered by `rm /tmp/ciao124`

GPL/Events/File/File.h
nicholasberlin: Combining after conversion works for me. Assuming it fits, of course (I'm not sure, else why the current separation of ns and sec)

GPL/Events/File/Probe.bpf.c
nicholasberlin: line 243 is the same situation.

nicholasberlin: and 425

GPL/Events/File/Probe.bpf.c
```diff
@@ -316,6 +342,7 @@ static int vfs_rename__enter(struct dentry *old_dentry, struct dentry *new_dentr
     ebpf_resolve_path_to_string(ss->rename.new_path, &p, task);
 
     state->rename.step = RENAME_STATE_PATHS_SET;
+    state->rename.de   = old_dentry;
```
nicholasberlin, line 345: Why not `new_dentry`? ... seems like we are reading from `rename.de` after `vfs_rename` has completed.

nicholasberlin (commented):
Ok, full pass done.

GPL/Events/File/File.h
mmat11: Go' syscall types do this -> https://pkg.go.dev/syscall#Timespec / https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/syscall/syscall.go;l=90

I added `atime` too

GPL/Events/File/Probe.bpf.c
mmat11: `new_dentry` is empty at that point; the data we are grabbing shouldn't change anyway, so both dentries should be fine, right?

GPL/Events/File/Probe.bpf.c
mmat11: added all file types handling

GPL/Events/File/Probe.bpf.c
stanek-michal: btw looking on Elixir at the newest kernel version - the arguments old_dentry and new_dentry were removed:
https://elixir.bootlin.com/linux/v6.5/source/fs/namei.c#L4754

stanek-michal: old_dentry looks fine here but note that `ctime` changes as a result of rename

mmat11: > btw looking on Elixir at the newest kernel version - the arguments old_dentry and new_dentry were removed:

this is handled here https://github.com/elastic/ebpf/blob/main/GPL/Events/File/Probe.bpf.c#L329

mmat11: > old_dentry looks fine here but note that ctime changes as a result of rename

new_dentry's inode is always empty both when using a pointer and when copying the struct in the state. If we need to get the `source` and `target` inodes, as they are called in the kernel source's vfs_rename, I think we need to hook another inner function... not sure it's worth it

nicholasberlin (approved):
LGTM

stanek-michal (approved):
LGTM

PR #185, opened 2024-03-19
Add probes for memfd_create and friends

stanek-michal:
New events added:

EBPF_EVENT_FILE_MEMFD_OPEN
EBPF_EVENT_FILE_SHMEM_OPEN
EBPF_EVENT_PROCESS_MEMFD_CREATE
EBPF_EVENT_PROCESS_SHMGET
EBPF_EVENT_PROCESS_PTRACE
EBPF_EVENT_PROCESS_LOAD_MODULE

Also added new fields to process exec event

### DIFF
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -18,13 +18,6 @@
 #include "vmlinux.h"
 #endif
 
-// from include/uapi/linux/posix_types.h
-/* Type of a SYSV IPC key. */
-typedef int key_t;
-
-// from include/uapi/asm-generic/posix_types.h
-typedef int pid_t;
-
 enum ebpf_event_type {
     EBPF_EVENT_PROCESS_FORK                 = (1 << 1),
     EBPF_EVENT_PROCESS_EXEC                 = (1 << 2),
@@ -236,7 +229,7 @@ struct ebpf_process_exec_event {
     struct ebpf_cred_info creds;
     struct ebpf_tty_dev ctty;
     char comm[TASK_COMM_LEN];
-    unsigned int inode_nlink;
+    uint32_t inode_nlink;
     uint32_t flags;
 
     // Variable length fields: cwd, argv, env, filename, pids_ss_cgroup_path
@@ -324,16 +317,16 @@ struct ebpf_process_memfd_create_event {
 struct ebpf_process_shmget_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
-    key_t key;
-    size_t size;
-    long shmflg; // shmget() flags
+    int32_t key;
+    uint64_t size;
+    int64_t shmflg; // shmget() flags
 } __attribute__((packed));
 
 struct ebpf_process_ptrace_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
-    pid_t child_pid;
-    long request;
+    int32_t child_pid;
+    int64_t request;
 } __attribute__((packed));
 
 struct ebpf_process_load_module_event {
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -226,9 +226,9 @@ struct ebpf_process_fork_event {
     struct ebpf_varlen_fields_start vl_fields;
 } __attribute__((packed));
 
-#define EXEC_F_SETUID	(1 << 0)
-#define EXEC_F_SETGID	(1 << 1)
-#define EXEC_F_MEMFD	(1 << 2)
+#define EXEC_F_SETUID   (1 << 0)
+#define EXEC_F_SETGID   (1 << 1)
+#define EXEC_F_MEMFD    (1 << 2)
 
 struct ebpf_process_exec_event {
     struct ebpf_event_header hdr;
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -18,6 +18,13 @@
 #include "vmlinux.h"
 #endif
 
+// from include/uapi/linux/posix_types.h
+/* Type of a SYSV IPC key. */
+typedef int key_t;
+
+// from include/uapi/asm-generic/posix_types.h
+typedef int pid_t;
+
 enum ebpf_event_type {
     EBPF_EVENT_PROCESS_FORK                 = (1 << 1),
     EBPF_EVENT_PROCESS_EXEC                 = (1 << 2),
@@ -30,9 +37,15 @@ enum ebpf_event_type {
     EBPF_EVENT_FILE_CREATE                  = (1 << 9),
     EBPF_EVENT_FILE_RENAME                  = (1 << 10),
     EBPF_EVENT_FILE_MODIFY                  = (1 << 11),
-    EBPF_EVENT_NETWORK_CONNECTION_ACCEPTED  = (1 << 12),
-    EBPF_EVENT_NETWORK_CONNECTION_ATTEMPTED = (1 << 13),
-    EBPF_EVENT_NETWORK_CONNECTION_CLOSED    = (1 << 14),
+    EBPF_EVENT_FILE_MEMFD_OPEN              = (1 << 12),
+    EBPF_EVENT_FILE_SHMEM_OPEN              = (1 << 13),
+    EBPF_EVENT_NETWORK_CONNECTION_ACCEPTED  = (1 << 14),
+    EBPF_EVENT_NETWORK_CONNECTION_ATTEMPTED = (1 << 15),
+    EBPF_EVENT_NETWORK_CONNECTION_CLOSED    = (1 << 16),
+    EBPF_EVENT_PROCESS_MEMFD_CREATE         = (1 << 17),
+    EBPF_EVENT_PROCESS_SHMGET               = (1 << 18),
+    EBPF_EVENT_PROCESS_PTRACE               = (1 << 19),
+    EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 20),
 };
 
 struct ebpf_event_header {
@@ -57,6 +70,8 @@ enum ebpf_varlen_field_type {
     EBPF_VL_FIELD_TTY_OUT,
     EBPF_VL_FIELD_PIDS_SS_CGROUP_PATH,
     EBPF_VL_FIELD_SYMLINK_TARGET_PATH,
+    EBPF_VL_FIELD_MOD_VERSION,
+    EBPF_VL_FIELD_MOD_SRCVERSION,
 };
 
 // Convenience macro to iterate all the variable length fields in an event
@@ -154,6 +169,7 @@ struct ebpf_file_delete_event {
     struct ebpf_varlen_fields_start vl_fields;
 } __attribute__((packed));
 
+// reused by memfd_open and shmem_open events
 struct ebpf_file_create_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
@@ -216,6 +232,10 @@ struct ebpf_process_exec_event {
     struct ebpf_cred_info creds;
     struct ebpf_tty_dev ctty;
     char comm[TASK_COMM_LEN];
+    unsigned int inode_nlink;
+    bool is_setuid;
+    bool is_setgid;
+    bool is_memfd;
 
     // Variable length fields: cwd, argv, env, filename, pids_ss_cgroup_path
     struct ebpf_varlen_fields_start vl_fields;
@@ -269,6 +289,42 @@ struct ebpf_process_setgid_event {
     uint32_t new_egid;
     uint32_t new_ruid;
     uint32_t new_euid;
+} __attribute__((packed));
+
+struct ebpf_process_memfd_create_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    unsigned int flags; // memfd_create flags
+    bool flag_cloexec;
+    bool flag_allow_seal;
+    bool flag_hugetlb;
+    bool flag_noexec_seal;
+    bool flag_exec;
+
+    // Variable length fields: memfd name
+    struct ebpf_varlen_fields_start vl_fields;
+} __attribute__((packed));
+
+struct ebpf_process_shmget_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    key_t key;
+    size_t size;
+    long shmflg; // shmget() flags
+} __attribute__((packed));
+
+struct ebpf_process_ptrace_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    pid_t child_pid;
+    long request;
+} __attribute__((packed));
+
+struct ebpf_process_load_module_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    // Variable length fields: filename, mod version, mod srcversion
+    struct ebpf_varlen_fields_start vl_fields;
 } __attribute__((packed));
 
 enum ebpf_net_info_transport {
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -226,6 +226,10 @@ struct ebpf_process_fork_event {
     struct ebpf_varlen_fields_start vl_fields;
 } __attribute__((packed));
 
+#define EXEC_F_SETUID	(1 << 0)
+#define EXEC_F_SETGID	(1 << 1)
+#define EXEC_F_MEMFD	(1 << 2)
+
 struct ebpf_process_exec_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
@@ -233,9 +237,7 @@ struct ebpf_process_exec_event {
     struct ebpf_tty_dev ctty;
     char comm[TASK_COMM_LEN];
     unsigned int inode_nlink;
-    bool is_setuid;
-    bool is_setgid;
-    bool is_memfd;
+    uint32_t flags;
 
     // Variable length fields: cwd, argv, env, filename, pids_ss_cgroup_path
     struct ebpf_varlen_fields_start vl_fields;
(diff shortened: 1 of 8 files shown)


GPL/Events/File/Probe.bpf.c
```diff
@@ -216,6 +216,60 @@ int BPF_KPROBE(kprobe__vfs_unlink)
...
+    ebpf_vl_field__set_size(&event->vl_fields, field, size);
+
+    // pids ss cgroup path
+    field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_PIDS_SS_CGROUP_PATH);
+    size  = ebpf_resolve_pids_ss_cgroup_path_to_string(field->data, task);
+    ebpf_vl_field__set_size(&event->vl_fields, field, size);
+
+    // skip event if prefix is specified and file path does not start with it
+    if (path_prefix) {
+        if ((path_prefix_len > 0) && (size >= path_prefix_len)) {
```
nicholasberlin, line 264: Why check `size >= path_prefix_len`? I think the current value in `size` is the cgroup path length.

GPL/Events/File/Probe.bpf.c
```diff
@@ -216,6 +216,60 @@ int BPF_KPROBE(kprobe__vfs_unlink)
...
+            if (is_equal_prefix(field->data, path_prefix, path_prefix_len))
```
nicholasberlin, line 265: Not sure about `field->data` here.

GPL/Events/Process/Probe.bpf.c
```diff
@@ -95,6 +98,30 @@ int BPF_PROG(sched_process_exec,
     ebpf_cred_info__fill(&event->creds, task);
     ebpf_ctty__fill(&event->ctty, task);
 
+    if (!binprm)
```
nicholasberlin, line 101: -- bikeshedding --
This if block coud be moved to the beginning of the function to avoid filling pid, cred, and ctty info.
Also, other guard logic uses `goto out`, which in this case is equivalent to `return 0`, but I think they should be consistent.

All that said, I'm guessing this will effectively always be false so there will never be any efficiency gains from moving it.
-- /bikeshedding --

GPL/Events/Process/Probe.bpf.c
```diff
@@ -222,6 +249,212 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
...
+    // Variable length fields
+    ebpf_vl_fields__init(&event->vl_fields);
+    struct ebpf_varlen_field *field;
+    long size;
+
+    // memfd filename
+    field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_FILENAME);
+    size  = bpf_probe_read_user_str(field->data, PATH_MAX, ex_args->uname);
+    if (size < 0)
+        return 1;
```
nicholasberlin, line 449: What does returning 1 do?

nicholasberlin (commented):
Nice work. A few questions.

GPL/Events/File/Probe.bpf.c
stanek-michal: thanks for spotting this, it's a rebase artifact, 'path' was supposed to be right above the path_prefix check

GPL/Events/Helpers.h
```diff
@@ -274,4 +278,15 @@ static bool is_consumer()
     return consumer_pid == pid;
 }
 
+// compares first 'len' characters of str1 and str2, returns 1 if equal
+static int is_equal_prefix(const char *str1, const char *str2, int len)
```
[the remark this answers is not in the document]
stanek-michal: I'm not 100% sure, but even if it didn't reject a case without fixed size arrays, it would probably just return 0 on a dereference if it was a bad page.

GPL/Events/Process/Probe.bpf.c
stanek-michal: might have been a typo, but either way as far as I know the return value is not considered

GPL/Events/Process/Probe.bpf.c
```diff
@@ -222,6 +249,212 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
...
+
+    // Variable length fields
+    ebpf_vl_fields__init(&event->vl_fields);
+    struct ebpf_varlen_field *field;
+    long size;
+
+    // memfd filename
+    field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_FILENAME);
+    size  = bpf_probe_read_user_str(field->data, PATH_MAX, ex_args->uname);
+    if (size < 0)
```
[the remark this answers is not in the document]
stanek-michal: better change to <= 0

nicholasberlin (approved):
LGTM if it works

mjwolf (approved):
The multi-kernel tester has a failing test with this, I didn't look into the exact cause, but this LGTM, once the tests are fixed

stanek-michal:
> The multi-kernel tester has a failing test with this, I didn't look into the exact cause, but this LGTM, once the tests are fixed

turned out it was a missing `break` in a switch block in EventsTrace that sometimes would trigger memory corruption and fail the multi-kernel tests. Fixed now and rebased, rerunning CI

GPL/Events/EbpfEventProto.h
```diff
@@ -216,6 +232,10 @@ struct ebpf_process_exec_event {
     struct ebpf_cred_info creds;
     struct ebpf_tty_dev ctty;
     char comm[TASK_COMM_LEN];
+    unsigned int inode_nlink;
+    bool is_setuid;
+    bool is_setgid;
```
haesbaert, line 237: Can we not use bool on things that show up on the wire, if you look at the rest of the file it uses properly sized types: uint32_t, int32_t and family, it even avoids using a single int.

GPL/Events/Process/Probe.bpf.c
```diff
@@ -97,6 +103,27 @@ int BPF_PROG(sched_process_exec,
     ebpf_ctty__fill(&event->ctty, task);
     ebpf_comm__fill(event->comm, sizeof(event->comm), task);
 
+    // set setuid and setgid flags
+    struct file *f        = BPF_CORE_READ(binprm, file);
+    struct inode *f_inode = BPF_CORE_READ(f, f_inode);
+    event->is_setuid      = (BPF_CORE_READ(f_inode, i_mode) & S_ISUID) ? true : false;
+    event->is_setgid      = (BPF_CORE_READ(f_inode, i_mode) & S_ISGID) ? true : false;
```
haesbaert, line 110: nitpick, `&` has precedence, I'd zap the parenthesis

haesbaert: you can also do:
`event->is_setuid = !!(BPF_CORE_READ(f_inode, i_mode) & S_ISUID)`

GPL/Events/Process/Probe.bpf.c
```diff
@@ -226,6 +253,212 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
...
+    if (is_kernel_thread(task))
+        goto out;
+
+    ebpf_pid_info__fill(&event->pids, task);
+
+    pid_t ppid      = BPF_CORE_READ(task, group_leader, real_parent, tgid);
+    pid_t curr_tgid = BPF_CORE_READ(task, tgid);
+
+    // ignore if process is child of init/systemd/whatever
+    if ((1 == curr_tgid) || (2 == curr_tgid) || (1 == ppid) || (2 == ppid))
```
haesbaert, line 280: can we avoid the 1 == curr_tgid, the rest of the repository uses curr_tgid == 1, I know this is endpoint leakage :D

haesbaert: can also zap the parenthesis, as == has precedence.

GPL/Events/Process/Probe.bpf.c
```diff
@@ -226,6 +253,212 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
...
+        char common_flags;
+        char common_preempt_count;
+        int common_pid;
+        int __syscall_nr;
+        long key;
+        size_t size;
+        long shmflg;
+    };
+    struct shmget_args *ex_args = (struct shmget_args *)ctx;
+
```
haesbaert, line 372: blank line

GPL/Events/Process/Probe.bpf.c
```diff
@@ -226,6 +253,212 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
...
+/* flags for memfd_create(2) (unsigned int) */
+#define MFD_CLOEXEC 0x0001U
+#define MFD_ALLOW_SEALING 0x0002U
+#define MFD_HUGETLB 0x0004U
+/* not executable and sealed to prevent changing to executable. */
+#define MFD_NOEXEC_SEAL 0x0008U
+/* executable */
+#define MFD_EXEC 0x0010U
+    event->flags            = ex_args->flags;
+    event->flag_cloexec     = (event->flags & MFD_CLOEXEC) ? true : false;
```
haesbaert, line 436: why do we have to express every flag isolated again in a member, can't the user just or the original flags?

GPL/Events/Helpers.h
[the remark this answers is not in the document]
haesbaert: I feel uneasy with this, and relying on the verifier solely seems not awesome, here is a valid alternative:
```
static int
strncmp(const char *s1, const char *s2, size_t n)
{

	if (n == 0)
		return (0);
	do {
		if (*s1 != *s2++)
			return (*(unsigned char *)s1 - *(unsigned char *)--s2);
		if (*s1++ == 0)
			break;
	} while (--n != 0);
	return (0);
}

// compares first 'len' characters of str1 and str2, returns 1 if equal
// NOTE: no bounds check, assumes use under eBPF verifier
static int is_equal_prefix(const char *str1, const char *str2, int len)
{
	return strncmp(str1, str2, len);
}

```

Compiles and runs with `EventsTrace --file-memfd-open --file-shmem-open --file-modify`

GPL/Events/EbpfEventProto.h
haesbaert: Can we condense the 3 members into a uin32_t and use a flag for each bit?

GPL/Events/EbpfEventProto.h
```diff
@@ -271,6 +291,42 @@ struct ebpf_process_setgid_event {
...
+struct ebpf_process_memfd_create_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    unsigned int flags; // memfd_create flags
+    bool flag_cloexec;
+    bool flag_allow_seal;
+    bool flag_hugetlb;
+    bool flag_noexec_seal;
+    bool flag_exec;
+
```
haesbaert, line 303: As noted somewhere else, why do we express flags again? Those 5 bits are in flags already, so why do we need to place them alone again?
I'm just eyeballing, so apologies if I missed something.

GPL/Events/EbpfEventProto.h
haesbaert: I've made this change that I think you can merge into this PR, my github-fu is severely limited, and as I said earlier, I ran out of goats:
https://github.com/elastic/ebpf/pull/194

haesbaert (changes requested):
.

GPL/Events/Helpers.h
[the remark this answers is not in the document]
stanek-michal: lgtm, I'll replace it

GPL/Events/EbpfEventProto.h
stanek-michal: agreed overall, there is an enum somewhere technically so it's not 100% portable but considering we want to be able to interface this from different languages, it's worth paying attention to

GPL/Events/Process/Probe.bpf.c
stanek-michal: that's one thing I do not agree with BSD folks, I like my parentheses and hate requiring everyone to know all precedence rules

GPL/Events/EbpfEventProto.h
haesbaert: Agreed, it's more for harmony with the rest than being portable, but it does help with stuff like someone using 
Go and having to figure out what is the size of bool.

GPL/Events/Process/Probe.bpf.c
haesbaert: Fine by me, please disregard the other comments regarding precedence.
Now I'm in an existential crisis, am I still BSD folk, not even god knows :rofl: .

GPL/Events/Process/Probe.bpf.c
```diff
@@ -226,6 +257,195 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
...
+    if (!event)
+        goto out;
+
+    event->hdr.type = EBPF_EVENT_PROCESS_LOAD_MODULE;
+    event->hdr.ts   = bpf_ktime_get_ns();
+
+    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+
+    if (is_kernel_thread(task))
+        goto out;
```
haesbaert, line 276:
```suggestion
    struct task_struct *task = (struct task_struct *)bpf_get_current_task();

    if (is_kernel_thread(task))
        goto out;

    struct ebpf_process_load_module_event *event = get_event_buffer();
    if (!event)
        goto out;

    event->hdr.type = EBPF_EVENT_PROCESS_LOAD_MODULE;
    event->hdr.ts   = bpf_ktime_get_ns();

```

Rest of the source checks task before getting event, I think we should keep the same pattern, pattern are goods and make for a healthy breakfast.

GPL/Events/Process/Probe.bpf.c
```diff
@@ -226,6 +257,195 @@ int tracepoint_syscalls_sys_exit_setsid(struct trace_event_raw_sys_exit *args)
...
+
+    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+    pid_t curr_tgid          = BPF_CORE_READ(task, tgid);
+    pid_t child_ppid         = BPF_CORE_READ(child, group_leader, real_parent, tgid);
+    pid_t child_tgid         = BPF_CORE_READ(child, tgid);
+
+    if (is_kernel_thread(task))
+        goto out;
+
```
haesbaert, line 333:
```suggestion
    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
    if (is_kernel_thread(task))
        goto out;

    pid_t curr_tgid          = BPF_CORE_READ(task, tgid);
    pid_t child_ppid         = BPF_CORE_READ(child, group_leader, real_parent, tgid);
    pid_t child_tgid         = BPF_CORE_READ(child, tgid);
```
Don't bother reading things if we're about to bail

cmake/modules/setup.cmake
```diff
@@ -65,6 +65,8 @@ function(ebpf_static_binary target)
 
     add_executable(${target} ${EBPF_BIN_SOURCES})
 
+    target_compile_options(${target} PUBLIC -g -Wall -Wextra)
```
haesbaert, line 68: does this apply to the whole repository?
I'm not very familiar with cmake, but I've added a bunch of flags in `non-GPL/CMakeLists.txt`, do they get merged?

haesbaert (changes requested):
Nice work, some very minor thingies

GPL/Events/Helpers.h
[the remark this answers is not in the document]
stanek-michal: added at the top

GPL/Events/EbpfEventProto.h
```diff
@@ -271,6 +286,56 @@ struct ebpf_process_setgid_event {
...
+    struct ebpf_pid_info pids;
+    int32_t key;
+    uint64_t size;
+    int64_t shmflg; // shmget() flags
+} __attribute__((packed));
+
+struct ebpf_process_ptrace_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    int32_t child_pid;
```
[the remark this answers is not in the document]
stanek-michal: done

GPL/Events/EbpfEventProto.h
```diff
@@ -271,6 +286,56 @@ struct ebpf_process_setgid_event {
...
+    struct ebpf_pid_info pids;
+    uint32_t flags; // memfd_create flags
+    // Variable length fields: memfd name
+    struct ebpf_varlen_fields_start vl_fields;
+} __attribute__((packed));
+
+struct ebpf_process_shmget_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    int32_t key;
```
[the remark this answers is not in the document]
stanek-michal: done

GPL/Events/Process/Probe.bpf.c
```diff
@@ -97,6 +103,31 @@ int BPF_PROG(sched_process_exec,
...
+        event->flags |= EXEC_F_SETGID;
+
+    // set inode link count (0 means anonymous or deleted file)
+    event->inode_nlink = BPF_CORE_READ(f_inode, i_nlink);
+
+    // check if memfd file is being exec'd
+    struct path p              = BPF_CORE_READ(binprm, file, f_path);
+    struct dentry *curr_dentry = BPF_CORE_READ(&p, dentry);
+    struct qstr component      = BPF_CORE_READ(curr_dentry, d_name);
+    char buf_filename[8]       = {0};
```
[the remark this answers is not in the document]
stanek-michal: done

GPL/Events/Process/Probe.bpf.c
stanek-michal: done

GPL/Events/Process/Probe.bpf.c
stanek-michal: done

cmake/modules/setup.cmake
stanek-michal: I think this only applies to calls to ebpf_static_binary, like this one:
`ebpf_static_binary(EventsTrace`
so EventsTrace will have debug symbols.

As for your change, not sure but best way is to look at the compilation output and find the flags

haesbaert (approved):

haesbaert:
Looking swell!

PR #190, opened 2024-05-02
Define the iterator in the for scope

haesbaert:
Otherwise running two FOREACHes in the same scope won't work as it tries to redefine the variable.


nicholasberlin (commented):
LGTM if the multi-kernel tests pass (which will run automagically.) This feels like a verifier work-around to me, but 🤷 

Also, looks like the linting gods require sacrifice.

nicholasberlin (approved):

PR #192, opened 2024-05-03
Add COMM to process events fork/exec/exit

haesbaert:
This is needed by quark and it makes sense at any rate since we don't track COMM changes explicitly (setproctitle and friends).


nicholasberlin (approved):

Tacklebox (approved):
LGTM

PR #194, opened 2024-05-22
Condense is_foo from exec into flags.

haesbaert:
Compiles and runs, needs proper testing and cleanup.


nicholasberlin (approved):
LGTM FWIW

stanek-michal (approved):

PR #195, opened 2024-06-03
Add ctty to fork

haesbaert:
Since we don't track actual TTY changes, at least make sure a new process gets the correct TTY, previously we were tracking it only on exec which is fine for probably most cases.


nicholasberlin (approved):

GPL/Events/EbpfEventProto.h
```diff
@@ -213,6 +213,7 @@ struct ebpf_process_fork_event {
     struct ebpf_pid_info parent_pids;
     struct ebpf_pid_info child_pids;
     struct ebpf_cred_info creds;
+    struct ebpf_tty_dev ctty;
```
stanek-michal, line 216: make sure to update ProcessForkEvent in utils.go to match

mjwolf (approved):

PR #198, opened 2024-07-02
Basic ringbuf event statistics

haesbaert:
This adds a new structure `ebpf_event_stats` where we accumulate stats, right now it only counts lost events vs non-lost events, but we can grow this in the future, like adding per type events if it's interesting enough.

The print in EventsTrace is quite primitive as I believe this will only be used for debugging there, so heh.

Choosing a name for `ebpf_ringbuf_write` was a bit tricky, I wanted to keep the same parameters but it had to be "different enough" that we could spot it with a naked eye.

While here, adjust the enum of ebpf events to not skip that precious first bit.

### DIFF
--- a/non-GPL/Events/EventsTrace/EventsTrace.c
+++ b/non-GPL/Events/EventsTrace/EventsTrace.c
@@ -35,7 +35,7 @@ const char argp_program_doc[] =
     "[--process-setgid] [--process-tty-write] [--process-memfd_create] [--process-shmget] "
     "[--process-ptrace] [--process-load_module]\n"
     "[--net-conn-accept] [--net-conn-attempt] [--net-conn-closed]\n"
-    "[--print-features-on-init] [--unbuffer-stdout] [--libbpf-verbose]\n";
+    "[--print-features-on-init] [--stats|-s] [--unbuffer-stdout] [--libbpf-verbose]\n";
 
 // Somewhat kludgy way of ensuring argp doesn't print the EBPF_* constants that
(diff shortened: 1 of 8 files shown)


non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -1179,6 +1184,14 @@ int main(int argc, char **argv)
             fprintf(stderr, "Failed to poll event context %d: %s\n", err, strerror(-err));
             break;
         }
+        if (g_stats) {
+            struct ebpf_event_stats ees;
+
+            if (ebpf_event_ctx__read_stats(ctx, &ees) == 0)
+                printf("sent %lu lost %lu\n", ees.sent, ees.lost);
```
nicholasberlin, line 1191: How often does this print?

haesbaert: It's awful, it's basically all the time non-stop because EventsTrace captures all TTY output, so it goes into a feedback loop.
I didn't care much since this is purely for debugging/testing, but I'm all for suggestions.

haesbaert: I mean, you still have to pass `-s`

nicholasberlin: Ok, makes sense.

nicholasberlin (approved):

mjwolf (approved):

Issue #200, opened 2024-08-23
`fexit/inet_csk_accept` probe will not work on kernel >=6.10


fearful-symmetry:
### Severity

Major

### Affects Version(s)

8.6.0

### Kernel Version(s)

6.10.3-200

### Linux Distro

Fedora 30

### Description

A commit 5 months ago changed the function signature of `inet_csk_accept`: https://github.com/torvalds/linux/commit/92ef0fd55ac80dfc2e4654edfe5d1ddfa6e070fe

On the kernel on my dev machine, it looks like this: `struct sock *inet_csk_accept(struct sock *sk, struct proto_accept_arg *arg)`

The code is not currently prepared for this possibility:
```
sudo ./EventsTrace                                                                                                                                                  1 ↵
libbpf: prog 'fexit__inet_csk_accept': BPF program load failed: Permission denied
libbpf: prog 'fexit__inet_csk_accept': -- BEGIN PROG LOAD LOG --
0: R1=ctx() R10=fp0
; int BPF_PROG( @ Probe.bpf.c:45
0: (79) r6 = *(u64 *)(r1 +32)
func 'inet_csk_accept' doesn't have 4-th argument
invalid bpf_context access off=32 size=8
processed 1 insns (limit 1000000) max_states_per_insn 0 total_states 0 peak_states 0 mark_read 0
-- END PROG LOAD LOG --
libbpf: prog 'fexit__inet_csk_accept': failed to load: -13
libbpf: failed to load object 'EventProbe_bpf'
libbpf: failed to load BPF skeleton 'EventProbe_bpf': -13
Could not create event context: -13 Permission denied
```

PR #201, opened 2024-08-26
Use `FUNC_RET_READ` for `inet_csk_accept`

fearful-symmetry:
closes https://github.com/elastic/ebpf/issues/200

This changes the `inet_csk_accept` fexit probe to use `FUNC_RET_READ` instead of passing args to the method; this means we don't have to deal with the function arguments changing, like with the 6.10 kernel.

Tested on Fedora with 6.10.3, but we seem to be missing the addr fields. Not sure if this is a quirk of using the fexit probe, or if there's a bug here and the `inet_csk_accept` method no longer works the way we assume.
```
sudo ./EventsTrace --net-conn-accept -i                                                                                                                                                   
{"probes_initialized": true, "features": {"bpf_tramp": true}}
{"event_type":"NETWORK_CONNECTION_ACCEPTED","pids":{"tid":4184470,"tgid":4184470,"ppid":4145903,"pgid":4184470,"sid":4145903,"start_time_ns":1468277566951535},"net":{"transport":"TCP","family":"AF_INET","source_address":"0.0.0.0","source_port":8000,"destination_address":"0.0.0.0","destination_port":0,"network_namespace":4026531840},"comm":"python3"}
```

### DIFF
--- a/GPL/Events/Helpers.h
+++ b/GPL/Events/Helpers.h
@@ -113,6 +113,7 @@ const volatile int consumer_pid = 0;
         ret;                                                                                       \
     })
 
+// value is replaced later by `probe_fill_relos()`
 #define DECL_FUNC_RET(func) const volatile int ret__##func##__ = 0;
 #define FUNC_RET_READ(type, func)                                                                  \
     ({                                                                                             \
--- a/GPL/Events/Network/Probe.bpf.c
+++ b/GPL/Events/Network/Probe.bpf.c
@@ -18,6 +18,8 @@
 #include "Network.h"
 #include "State.h"
 
+DECL_FUNC_RET(inet_csk_accept);
+
 static int inet_csk_accept__exit(struct sock *sk)
 {
     if (!sk)
@@ -42,9 +44,9 @@ out:
 }
 
 SEC("fexit/inet_csk_accept")
-int BPF_PROG(
-    fexit__inet_csk_accept, struct sock *sk, int flags, int *err, bool kern, struct sock *ret)
+int BPF_PROG(fexit__inet_csk_accept)
 {
+    struct sock *ret = FUNC_RET_READ(___type(ret), inet_csk_accept);
     return inet_csk_accept__exit(ret);
 }
 
--- a/non-GPL/Events/Lib/EbpfEvents.c
+++ b/non-GPL/Events/Lib/EbpfEvents.c
@@ -268,7 +268,7 @@ static int probe_fill_relos(struct btf *btf, struct EventProbe_bpf *obj)
 {
     int err = 0;
 
-    err = err ?: FILL_FUNC_RET_IDX(obj,btf, inet_csk_accept);
+    err = err ?: FILL_FUNC_RET_IDX(obj, btf, inet_csk_accept);
 
     err = err ?: FILL_FUNC_ARG_IDX(obj, btf, vfs_unlink, dentry);
     err = err ?: FILL_FUNC_RET_IDX(obj, btf, vfs_unlink);
--- a/non-GPL/Events/Lib/EbpfEvents.c
+++ b/non-GPL/Events/Lib/EbpfEvents.c
@@ -268,6 +268,8 @@ static int probe_fill_relos(struct btf *btf, struct EventProbe_bpf *obj)
 {
     int err = 0;
 
+    err = err ?: FILL_FUNC_RET_IDX(obj,btf, inet_csk_accept);
+
     err = err ?: FILL_FUNC_ARG_IDX(obj, btf, vfs_unlink, dentry);
     err = err ?: FILL_FUNC_RET_IDX(obj, btf, vfs_unlink);
 


haesbaert:
Not sure I grok this but, won't this break for multiple cpus? Two cpus entering the same function would basically store the return value in the same storage space.
I guess the magic is in the `ctx` arithmetic, where does that come from?

fearful-symmetry:
@haesbaert hmm, good point, didn't think of that. @stanek-michal do you know how this will work across CPUs?

fearful-symmetry:
Using the branch where I rewrote the probes to use an updated vmlinux, it works fine:
```
sudo ./EventsTrace --net-conn-accept -i                                                                                                              
{"probes_initialized": true, "features": {"bpf_tramp": true}}
{"event_type":"NETWORK_CONNECTION_ACCEPTED","pids":{"tid":4184470,"tgid":4184470,"ppid":4145903,"pgid":4184470,"sid":4145903,"start_time_ns":1468277566951535},"net":{"transport":"TCP","family":"AF_INET","source_address":"192.168.1.37","source_port":8000,"destination_address":"192.168.1.124","destination_port":54728,"network_namespace":4026531840},"comm":"python3"}
``` 
I guess `FUNC_RET_READ` is behaving in some unexpected way here?

haesbaert:
So I was wrong, ctx + 0 is supposed to be where the return address is, but I can't seem to find the layout of ctx.

haesbaert:
with the rellocation addition it's correct I believe.

stanek-michal:
> So I was wrong, ctx + 0 is supposed to be where the return address is, but I can't seem to find the layout of ctx.

Yeah, its eBPF's R0 which is the return value, the layout I guess is just the eBPF registers (like pt_regs but for eBPF arch)

stanek-michal (approved):
Looks good, I don't see any issues, multiple CPUs don't seem to matter here. Thanks for the changes!

haesbaert:
> > So I was wrong, ctx + 0 is supposed to be where the return address is, but I can't seem to find the layout of ctx.
> 
> Yeah, its eBPF's R0 which is the return value, the layout I guess is just the eBPF registers (like pt_regs but for eBPF arch)

Not sure that's quite right, if I understand this now, it does `ctx + patched_return_index`, which is the last register in the return function:

```
int BPF_PROG(
    fexit__inet_csk_accept, struct sock *sk, int flags, int *err, bool kern, struct sock *ret)
```

So the patched_return_index here would be 4, meaning use R5 (R1=sk; R2=flags; R3=err; R4=kern; R5=ret).
Problem before is we were doing ctx + 0 which is R1.

stanek-michal:
> > > So I was wrong, ctx + 0 is supposed to be where the return address is, but I can't seem to find the layout of ctx.
> > 
> > 
> > Yeah, its eBPF's R0 which is the return value, the layout I guess is just the eBPF registers (like pt_regs but for eBPF arch)
> 
> Not sure that's quite right, if I understand this now, it does `ctx + patched_return_index`, which is the last register in the return function:
> 
> ```
> int BPF_PROG(
>     fexit__inet_csk_accept, struct sock *sk, int flags, int *err, bool kern, struct sock *ret)
> ```
> 
> So the patched_return_index here would be 4, meaning use R5 (R1=sk; R2=flags; R3=err; R4=kern; R5=ret). Problem before is we were doing ctx + 0 which is R1.

Oh interesting. I guess the ctx (which seems to be struct 'bpf_context') can be different things in different cases, like for network stuff its the packet data etc.

fearful-symmetry:
Not really sure what's going on with the CI failures, can't see anything other than this:

```
make: *** [Makefile:171: run-multikernel-test] Error 1
```

haesbaert:
> > > > So I was wrong, ctx + 0 is supposed to be where the return address is, but I can't seem to find the layout of ctx.
> > > 
> > > 
> > > Yeah, its eBPF's R0 which is the return value, the layout I guess is just the eBPF registers (like pt_regs but for eBPF arch)
> > 
> > 
> > Not sure that's quite right, if I understand this now, it does `ctx + patched_return_index`, which is the last register in the return function:
> > ```
> > int BPF_PROG(
> >     fexit__inet_csk_accept, struct sock *sk, int flags, int *err, bool kern, struct sock *ret)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So the patched_return_index here would be 4, meaning use R5 (R1=sk; R2=flags; R3=err; R4=kern; R5=ret). Problem before is we were doing ctx + 0 which is R1.
> 
> Oh interesting. I guess the ctx (which seems to be struct 'bpf_context') can be different things in different cases, like for network stuff its the packet data etc.

I guess so as well, I'll try to read the bpf code later and see if I can confirm.

fearful-symmetry:
are the tests just....timing out?
```
2024-08-27T17:42:18.7669228Z [            ] stdout: timed out waiting for EventsTrace to get ready, dumped stderr above
2024-08-27T17:42:18.7669408Z [            ] stdout: ===== STACKTRACE FOR FAILED TEST =====
2024-08-27T17:42:18.7669543Z [            ] stdout: goroutine 1 [running]:
2024-08-27T17:42:18.7669709Z [            ] stdout: main.TestFail({0x400088fd98, 0x1, 0x1})
2024-08-27T17:42:18.7670123Z [            ] stdout: 	/home/runner/work/ebpf/ebpf/testing/testrunner/utils.go:292 +0xe8
2024-08-27T17:42:18.7670417Z [            ] stdout: main.(*EventsTraceInstance).Start(0x4000012080, {0x13d168, 0x400008c240})
2024-08-27T17:42:18.7670761Z [            ] stdout: 	/home/runner/work/ebpf/ebpf/testing/testrunner/eventstrace.go:87 +0x360
2024-08-27T17:42:18.7670983Z [            ] stdout: main.RunEventsTest(0x11dd40, {0x400009df38, 0x1, 0x1})
2024-08-27T17:42:18.7671283Z [            ] stdout: 	/home/runner/work/ebpf/ebpf/testing/testrunner/utils.go:350 +0x190
2024-08-27T17:42:18.7671395Z [            ] stdout: main.main()
2024-08-27T17:42:18.7671684Z [            ] stdout: 	/home/runner/work/ebpf/ebpf/testing/testrunner/main.go:30 +0x2dc
```

PR #204, opened 2024-08-27
Remove the option to print all events in EventsTrace

stanek-michal:
This option makes no sense because it spams tty events in an infinite feedback loop.


haesbaert:
can't you just remove tty from -a? cause it would be useful otherwise.

stanek-michal:
> can't you just remove tty from -a? cause it would be useful otherwise.

yeah seems better - added it

nicholasberlin (approved):

PR #206, opened 2024-08-30
Add UDP probes in anticipation of DNS monitoring

fearful-symmetry:
see: https://datatracker.ietf.org/doc/html/rfc1035#autoid-40

This PR adds basic support for DNS monitoring, using the `ip_send_skb` and `skb_consume_udp` functions in the kernel networking stack. The logic here is very basic, and includes no parsing of actual DNS data; after spending the better part of the day reading RFCs and looking at other implementations, I've decided that parsing actual DNS requests/responses is complex enough (and risky enough) that we should probably use an [external library](https://github.com/c-ares/c-ares/blob/main/include/ares_dns_record.h#L1087) for parsing DNS data. It's one of those "getting it 80% working is 20% of the code" protocols. 

As a result of this, this PR just creates a DNS packet that just provides basic network metadata and and a  buffer for the packet data.

### DIFF
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -11,6 +11,8 @@
 #define EBPF_EVENTPROBE_EBPFEVENTPROTO_H
 
 #define TASK_COMM_LEN 16
+#define MAX_DNS_PACKET 512
+#define MAX_NR_SEGS 8
 
 #ifndef __KERNEL__
 #include <stdint.h>
@@ -42,6 +44,7 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
     EBPF_EVENT_NETWORK_UDP_SENDMSG          = (1 << 20),
     EBPF_EVENT_NETWORK_UDP_RECVMSG          = (1 << 21),
+    EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 22),
 };
 
 struct ebpf_event_header {
@@ -380,6 +383,29 @@ struct ebpf_net_event {
     struct ebpf_pid_info pids;
     struct ebpf_net_info net;
     char comm[TASK_COMM_LEN];
+} __attribute__((packed));
+
+struct dns_pkt_header {
+    uint16_t transaction_id;
+    uint16_t flags;
+    uint16_t num_questions;
+    uint16_t num_answers;
+    uint16_t num_auth_rrs;
+    uint16_t num_additional_rrs;
+} __attribute__((packed));
+
+struct dns_body {
+    size_t len;
+    uint8_t pkt[MAX_DNS_PACKET];
+} __attribute((packed));
+
+struct ebpf_dns_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    struct ebpf_net_info net;
+    char comm[TASK_COMM_LEN];
+    enum ebpf_event_type udp_evt;
+    struct dns_body pkts[MAX_NR_SEGS];
 } __attribute__((packed));
 
 // Basic event statistics
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -42,8 +42,8 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
-    EBPF_EVENT_NETWORK_UDP_SENDMSG          = (1 << 20),
-    EBPF_EVENT_NETWORK_UDP_RECVMSG          = (1 << 21),
+    EBPF_EVENT_NETWORK_UDP_SEND          = (1 << 20),
+    EBPF_EVENT_NETWORK_UDP_RECV          = (1 << 21),
     EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 22),
 };
 
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -46,7 +46,7 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
-    EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 22),
+    EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 20),
 };
 
 struct ebpf_event_header {
@@ -393,11 +393,6 @@ struct ebpf_net_event {
     char comm[TASK_COMM_LEN];
 } __attribute__((packed));
 
-struct dns_body {
-    size_t len;
-    uint8_t pkt[MAX_DNS_PACKET];
-} __attribute((packed));
-
 struct ebpf_dns_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
@@ -406,7 +401,6 @@ struct ebpf_dns_event {
     enum ebpf_net_udp_info udp_evt;
     // Variable length fields: dns body
     struct ebpf_varlen_fields_start vl_fields;
-    // uint8_t pkt[MAX_DNS_PACKET];
 } __attribute__((packed));
 
 // Basic event statistics
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -11,12 +11,10 @@
 #define EBPF_EVENTPROBE_EBPFEVENTPROTO_H
 
 #define TASK_COMM_LEN 16
-// The theoretical max size of DNS packets over UDP
+// The theoretical max size of DNS packets over UDP is 512.
 // Like so many things in DNS this number probaby isn't 100% accurate.
-// DNS extensions in RFC2671 and RFC6891 mean the actual size can be larger,
-// although an additonal body over 512 is probably due to additional RR fields,
-// which we can (probably) get away with throwing out for now.
-#define MAX_DNS_PACKET 512
+// DNS extensions in RFC2671 and RFC6891 mean the actual size can be larger.
+#define MAX_DNS_PACKET 1500
 
 #ifndef __KERNEL__
 #include <stdint.h>
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -404,9 +404,10 @@ struct ebpf_dns_event {
 
 // Basic event statistics
 struct ebpf_event_stats {
-    uint64_t lost;      // lost events due to a full ringbuffer
-    uint64_t sent;      // events sent through the ringbuffer
-    uint64_t paged_out; // indicates that that a given sk_buff event contained no non-paged data
+    uint64_t lost; // lost events due to a full ringbuffer
+    uint64_t sent; // events sent through the ringbuffer
+    uint64_t
+        dns_body_paged_out; // indicates that an sk_buff for a DNS event contained no non-paged data
 };
 
 #endif // EBPF_EVENTPROBE_EBPFEVENTPROTO_H
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -42,8 +42,8 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
-    EBPF_EVENT_NETWORK_UDP_SEND          = (1 << 20),
-    EBPF_EVENT_NETWORK_UDP_RECV          = (1 << 21),
+    EBPF_EVENT_NETWORK_UDP_SEND             = (1 << 20),
+    EBPF_EVENT_NETWORK_UDP_RECV             = (1 << 21),
     EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 22),
 };
 
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -406,6 +406,7 @@ struct ebpf_dns_event {
 struct ebpf_event_stats {
     uint64_t lost; // lost events due to a full ringbuffer
     uint64_t sent; // events sent through the ringbuffer
+    uint64_t paged_out; // indicates that that a given sk_buff event contained no non-paged data
 };
 
 #endif // EBPF_EVENTPROBE_EBPFEVENTPROTO_H
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -42,8 +42,8 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
-    EBPF_EVENT_NETWORK_SEND_SKB             = (1 << 20),
-    EBPF_EVENT_NETWORK_CONSUME_SKB          = (1 << 21),
+    EBPF_EVENT_NETWORK_UDP_SENDMSG          = (1 << 20),
+    EBPF_EVENT_NETWORK_UDP_RECVMSG          = (1 << 21),
     EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 22),
 };
 
@@ -394,13 +394,18 @@ struct dns_pkt_header {
     uint16_t num_additional_rrs;
 } __attribute__((packed));
 
+struct dns_body {
+    size_t len;
+    uint8_t pkt[MAX_DNS_PACKET];
+} __attribute((packed));
+
 struct ebpf_dns_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
     struct ebpf_net_info net;
     char comm[TASK_COMM_LEN];
     enum ebpf_event_type udp_evt;
-    uint8_t pkt[MAX_DNS_PACKET];
+    struct dns_body pkts[MAX_NR_SEGS];
 } __attribute__((packed));
 
 // Basic event statistics
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -42,8 +42,8 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
-    EBPF_EVENT_NETWORK_UDP_SENDMSG          = (1 << 20),
-    EBPF_EVENT_NETWORK_UDP_RECVMSG          = (1 << 21),
+    EBPF_EVENT_NETWORK_SEND_SKB             = (1 << 20),
+    EBPF_EVENT_NETWORK_CONSUME_SKB          = (1 << 21),
     EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 22),
 };
 
@@ -394,18 +394,13 @@ struct dns_pkt_header {
     uint16_t num_additional_rrs;
 } __attribute__((packed));
 
-struct dns_body {
-    size_t len;
-    uint8_t pkt[MAX_DNS_PACKET];
-} __attribute((packed));
-
 struct ebpf_dns_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
     struct ebpf_net_info net;
     char comm[TASK_COMM_LEN];
     enum ebpf_event_type udp_evt;
-    struct dns_body pkts[MAX_NR_SEGS];
+    uint8_t pkt[MAX_DNS_PACKET];
 } __attribute__((packed));
 
 // Basic event statistics
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -406,8 +406,8 @@ struct ebpf_dns_event {
 struct ebpf_event_stats {
     uint64_t lost; // lost events due to a full ringbuffer
     uint64_t sent; // events sent through the ringbuffer
-    uint64_t
-        dns_body_paged_out; // indicates that an sk_buff for a DNS event contained no non-paged data
+    // indicates that an sk_buff for a DNS event contained no non-paged data in sk_buff->data
+    uint64_t dns_zero_body;
 };
 
 #endif // EBPF_EVENTPROBE_EBPFEVENTPROTO_H
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -397,6 +397,7 @@ struct ebpf_dns_event {
     struct ebpf_net_info net;
     char comm[TASK_COMM_LEN];
     enum ebpf_net_udp_info udp_evt;
+    uint64_t original_len;
     // Variable length fields: dns body
     struct ebpf_varlen_fields_start vl_fields;
 } __attribute__((packed));
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -404,8 +404,8 @@ struct ebpf_dns_event {
 
 // Basic event statistics
 struct ebpf_event_stats {
-    uint64_t lost; // lost events due to a full ringbuffer
-    uint64_t sent; // events sent through the ringbuffer
+    uint64_t lost;      // lost events due to a full ringbuffer
+    uint64_t sent;      // events sent through the ringbuffer
     uint64_t paged_out; // indicates that that a given sk_buff event contained no non-paged data
 };
 
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -12,7 +12,6 @@
 
 #define TASK_COMM_LEN 16
 #define MAX_DNS_PACKET 512
-#define MAX_NR_SEGS 8
 
 #ifndef __KERNEL__
 #include <stdint.h>
@@ -42,8 +41,6 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
-    EBPF_EVENT_NETWORK_UDP_SEND             = (1 << 20),
-    EBPF_EVENT_NETWORK_UDP_RECV             = (1 << 21),
     EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 22),
 };
 
@@ -354,6 +351,11 @@ enum ebpf_net_info_af {
     EBPF_NETWORK_EVENT_AF_INET6 = 2,
 };
 
+enum ebpf_net_udp_info {
+    EBPF_NETWORK_EVENT_SKB_CONSUME_UDP = 1,
+    EBPF_NETWORK_EVENT_IP_SEND_UDP     = 2,
+};
+
 struct ebpf_net_info_tcp_close {
     uint64_t bytes_sent;
     uint64_t bytes_received;
@@ -385,42 +387,18 @@ struct ebpf_net_event {
     char comm[TASK_COMM_LEN];
 } __attribute__((packed));
 
-struct dns_pkt_header {
-    uint16_t transaction_id;
-    uint16_t flags;
-    uint16_t num_questions;
-    uint16_t num_answers;
-    uint16_t num_auth_rrs;
-    uint16_t num_additional_rrs;
-} __attribute__((packed));
-
 struct dns_body {
     size_t len;
     uint8_t pkt[MAX_DNS_PACKET];
 } __attribute((packed));
-
-// from vmlinux.h, modified to make the ip addrs u8 arrays
-struct skb_iphdr {
-    uint8_t ihl : 4;
-    uint8_t version : 4;
-    uint8_t tos;
-    uint16_t tot_len;
-    uint16_t id;
-    uint16_t frag_off;
-    uint8_t ttl;
-    uint8_t protocol;
-    uint16_t check;
-    uint8_t saddr[4];
-    uint8_t daddr[4];
-};
 
 struct ebpf_dns_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
     struct ebpf_net_info net;
     char comm[TASK_COMM_LEN];
-    enum ebpf_event_type udp_evt;
-    struct dns_body pkts[MAX_NR_SEGS];
+    enum ebpf_net_udp_info udp_evt;
+    uint8_t pkt[MAX_DNS_PACKET];
 } __attribute__((packed));
 
 // Basic event statistics
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -40,8 +40,8 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
-    EBPF_EVENT_NETWORK_UDP_SENDMSG = (1 << 20),
-    EBPF_EVENT_NETWORK_UDP_RECVMSG = (1 << 21),
+    EBPF_EVENT_NETWORK_UDP_SENDMSG          = (1 << 20),
+    EBPF_EVENT_NETWORK_UDP_RECVMSG          = (1 << 21),
 };
 
 struct ebpf_event_header {
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -11,6 +11,11 @@
 #define EBPF_EVENTPROBE_EBPFEVENTPROTO_H
 
 #define TASK_COMM_LEN 16
+// The theoretical max size of DNS packets over UDP
+// Like so many things in DNS this number probaby isn't 100% accurate.
+// DNS extensions in RFC2671 and RFC6891 mean the actual size can be larger,
+// although an additonal body over 512 is probably due to additional RR fields,
+// which we can (probably) get away with throwing out for now.
 #define MAX_DNS_PACKET 512
 
 #ifndef __KERNEL__
@@ -68,6 +73,7 @@ enum ebpf_varlen_field_type {
     EBPF_VL_FIELD_SYMLINK_TARGET_PATH,
     EBPF_VL_FIELD_MOD_VERSION,
     EBPF_VL_FIELD_MOD_SRCVERSION,
+    EBPF_VL_FIELD_DNS_BODY,
 };
 
 // Convenience macro to iterate all the variable length fields in an event
@@ -398,7 +404,9 @@ struct ebpf_dns_event {
     struct ebpf_net_info net;
     char comm[TASK_COMM_LEN];
     enum ebpf_net_udp_info udp_evt;
-    uint8_t pkt[MAX_DNS_PACKET];
+    // Variable length fields: dns body
+    struct ebpf_varlen_fields_start vl_fields;
+    // uint8_t pkt[MAX_DNS_PACKET];
 } __attribute__((packed));
 
 // Basic event statistics
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -399,6 +399,21 @@ struct dns_body {
     uint8_t pkt[MAX_DNS_PACKET];
 } __attribute((packed));
 
+// from vmlinux.h, modified to make the ip addrs u8 arrays
+struct skb_iphdr {
+    uint8_t ihl : 4;
+    uint8_t version : 4;
+    uint8_t tos;
+    uint16_t tot_len;
+    uint16_t id;
+    uint16_t frag_off;
+    uint8_t ttl;
+    uint8_t protocol;
+    uint16_t check;
+    uint8_t saddr[4];
+    uint8_t daddr[4];
+};
+
 struct ebpf_dns_event {
     struct ebpf_event_header hdr;
     struct ebpf_pid_info pids;
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -40,6 +40,8 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
+    EBPF_EVENT_NETWORK_UDP_SENDMSG = (1 << 20),
+    EBPF_EVENT_NETWORK_UDP_RECVMSG = (1 << 21),
 };
 
 struct ebpf_event_header {
@@ -341,6 +343,7 @@ struct ebpf_process_load_module_event {
 
 enum ebpf_net_info_transport {
     EBPF_NETWORK_EVENT_TRANSPORT_TCP = 1,
+    EBPF_NETWORK_EVENT_TRANSPORT_UDP = 2,
 };
 
 enum ebpf_net_info_af {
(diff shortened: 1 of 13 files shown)


stanek-michal:
> using fentry/fexit this is (in theory) possible by reading the `iov_iter` struct. Where this gets complicated is pre-trampoline kernels---`kretprobe` doesn't have access to the function args, meaning we can't get the iovec data after it's been populated.

We could probably do the set_state/get_state trick that we have in Helpers.h which maintains state between entry and ret so that we get access to the input argument pointers in the handler for ret.

> There's some alternatives here, most notably using TC probes: https://ebpf-docs.dylanreimerink.nl/linux/program-type/BPF_PROG_TYPE_SCHED_CLS/ , which I haven't experimented with yet.

I believe this is where TC runs on the RX path:
https://elixir.bootlin.com/linux/v6.10.7/source/net/core/dev.c#L4049
so this is early, at L2, before `deliver_skb()` and before it reaches the IP layer. We probably cannot easily map it to the receiving process without having to maintain our own bookkeeping. We could experiment with [bpf_sk_lookup_tcp](https://patchwork.ozlabs.org/project/netdev/patch/20180921171043.20823-8-joe@wand.net.nz/) helper and see if its reliable - if it works as advertised it can get the listening socket just from the 5tuple so that could allow us to extract the PID easier probably. There's also `bpf_sk_fullsock()` which does...something.

Another thing is at this early L2 stage we don't even know if the packet will actually reach the application, there is quite a lot of filtering and drop scenarios in between.

So I think I'm leaning towards the hooks as they are done in this PR if we can easily get packet data there and parse it. I assume we want to filter for port 53 in the kernel and avoid creating events for other UDP packets otherwise there could be a lot.

fearful-symmetry:
@stanek-michal so, I did some tinkering over the three-day weekend and came to a lot of the conclusions you did, namely that we could use a state getter/setter for kretprobes, and also the problem with finding the userspace pid with a TC/classifier probe.


I'm leaning towards the getter/setter with kretprobe idea for this, since it's a little simpler (we don't need some other mechanism for tracking what userland process owns the pid).

fearful-symmetry:
Taking this out of draft since I think this is a fairly decent framework going forward if/when we want to parse DNS packets.

haesbaert:
This won't work for unconnected sockets, as the socket doesn't have an address, if you run the following program you will see the packets don't show up. Worth noting that unconnected udp sockets are likely more common than unconnected.

You can use the following program to verify:
```
./udpsend -c 8.8.8.8 # connected
./udpsend 8.8.8.8 # unconnected
```

```
#define _GNU_SOURCE 		/* program_invocation_name */
#include <sys/types.h>
#include <sys/socket.h>

#include <netinet/in.h>

#include <arpa/inet.h>

#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <errno.h>
#include <err.h>

static void
usage(void)
{
	fprintf(stderr, "usage: %s [-c] addr4",
	    program_invocation_short_name);

	exit(1);
}

int
main(int argc, char *argv[])
{
	struct sockaddr_in	sin;
	int			sock, ch, do_connect;
	ssize_t			n;
	uint64_t		buf[] = {
		0xdeadbeef, 0xdeadbeef, 0xdeadbeef, 0xdeadbeef,
		0xdeadbeef, 0xdeadbeef, 0xdeadbeef, 0xdeadbeef
	};

	do_connect = 0;
	while ((ch = getopt(argc, argv, "ch")) != -1) {
		switch (ch) {
		case 'c':
			do_connect = 1;
			break;
		default:
			usage();
		}
	}
	argc -= optind;
	argv += optind;

	bzero(&sin, sizeof(sin));

	if (inet_aton(argv[0], &sin.sin_addr) == 0)
		errx(1, "inet_aton(%s)", argv[0]);
	sin.sin_port = htons(53);
	sin.sin_family = AF_INET;

	if ((sock = socket(AF_INET, SOCK_DGRAM, 0)) == -1)
		err(1, "socket");

	if (do_connect) {
		if (connect(sock, (struct sockaddr *)&sin, sizeof(sin)) == -1)
			err(1, "connect");
		n = write(sock, buf, sizeof(buf));
		if (n == -1)
			err(1, "sendto");
		else if (n != sizeof(buf))
			errx(1, "sendto: shortcount");

		return (0);
	}

	n = sendto(sock, buf, sizeof(buf), 0,
	    (struct sockaddr *)&sin, sizeof(sin));
	if (n == -1)
		err(1, "sendto");
	else if (n != sizeof(buf))
		errx(1, "sendto: shortcount");

	return (0);
}

```

As I mentioned to you yesterday, I think this is still not a good place to tap, I'll do a proper review later tonight and have more comments, but if you use the program above with events trace you can spot the issue.

non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -1051,6 +1059,26 @@ static void out_network_connection_accepted_event(struct ebpf_net_event *evt)
     out_network_event("NETWORK_CONNECTION_ACCEPTED", evt);
 }
 
+static void out_network_udp_sendmsg(struct ebpf_net_event *evnt)
```
nicholasberlin, line 1062: delete this function?

non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -1051,6 +1059,26 @@ static void out_network_connection_accepted_event(struct ebpf_net_event *evt)
...
+{
+    out_network_event("NETWORK_UDP_SENDMSG", evnt);
+}
+
+static void out_network_udp_recvmsg(struct ebpf_net_event *evnt)
```
nicholasberlin, line 1067: delete this function?

GPL/Events/Network/Network.h
```diff
@@ -14,6 +14,14 @@
 #define AF_INET 2
 #define AF_INET6 10
 
+#define MSG_PEEK 2
```
nicholasberlin, line 17: I don't think these defines are used anymore.

GPL/Events/EbpfEventProto.h
```diff
@@ -379,6 +385,24 @@ struct ebpf_net_event {
...
+    uint16_t num_additional_rrs;
+} __attribute__((packed));
+
+struct ebpf_dns_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    struct ebpf_net_info net;
+    char comm[TASK_COMM_LEN];
+    enum ebpf_event_type udp_evt;
+    uint8_t pkt[MAX_DNS_PACKET];
```
nicholasberlin, line 403: Will userspace be able to figure out if the entire array is not used? Do we need a packet length field?

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,135 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    bpf_ringbuf_discard(event, 0);
+    return 0;
+}
+
+/*
+=============================== DNS probes ===============================
+*/
+
+// SEC("fentry/udp_sendmsg")
+// int BPF_PROG(fentry__udp_sendmsg, struct sock *sk, struct msghdr *msg, size_t size)
```
nicholasberlin, line 111: delete?

non-GPL/Events/EventsTrace/EventsTrace.c
nicholasberlin: oh, nevermind. looks like it should be renamed though.

non-GPL/Events/EventsTrace/EventsTrace.c
nicholasberlin: oh, nevermind. looks like it should be renamed though.

nicholasberlin (commented):
Looking good, just a bit of clean up left I think. Oh, and the tests are not passing.

fearful-symmetry:
Alright, so, reverted the commit that swapped over to the `*skb*` methods. This is back to using `udp_sendmsg` and `udp_recvmsg`. Will continue to investigate alternatives.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    u16 transport_header_offset = BPF_CORE_READ(skb, transport_header);
+
+    u8 iphdr_first_byte = 0;
+    bpf_core_read(&iphdr_first_byte, 1, data + net_header_offset);
+    iphdr_first_byte = iphdr_first_byte >> 4;
+
+    u8 proto = 0;
+    if (iphdr_first_byte == 4) {
+        struct iphdr ip_hdr;
+        bpf_core_read(&ip_hdr, sizeof(struct iphdr), data + net_header_offset);
```
haesbaert, line 73: This breaks endianess, the information you want is in iphdr:
```
struct iphdr {
#if defined(__LITTLE_ENDIAN_BITFIELD)
    __u8 ihl : 4, version : 4;
#elif defined(__BIG_ENDIAN_BITFIELD)
    __u8 version : 4, ihl : 4;
#else
#error "Please fix <asm/byteorder.h>"
#endif
    __u8 tos;
    __be16 tot_len;
    __be16 id;
    __be16 frag_off;
    __u8 ttl;
    __u8 protocol;
    __sum16 check;
    __be32 saddr;
    __be32 daddr;
    /*The options start here. */
};
```

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+
+        proto = ip_hdr.protocol;
+        bpf_probe_read(event->net.saddr, 4, (void *)&ip_hdr.saddr);
+        bpf_probe_read(event->net.daddr, 4, (void *)&ip_hdr.daddr);
+
```
haesbaert, line 78: Don't need to cast to `void *`

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    } else if (iphdr_first_byte == 6) {
+        struct ipv6hdr ip6_hdr;
+        bpf_core_read(&ip6_hdr, sizeof(struct ipv6hdr), data + net_header_offset);
+        proto = ip6_hdr.nexthdr;
+
+        bpf_probe_read(event->net.saddr6, 16, ip6_hdr.saddr.in6_u.u6_addr8);
+        bpf_probe_read(event->net.daddr6, 16, ip6_hdr.daddr.in6_u.u6_addr8);
```
haesbaert, line 85: I'm ignoring ipv6 for now.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    event->net.sport = bpf_ntohs(udp_hdr.source);
+
+    struct task_struct *task = (struct task_struct *)bpf_get_current_task();
+    ebpf_pid_info__fill(&event->pids, task);
+    bpf_get_current_comm(event->comm, TASK_COMM_LEN);
+    event->hdr.ts = bpf_ktime_get_ns();
+
+    // filter out non-dns packets
+    if (event->net.sport != 53 && event->net.dport != 53) {
+        bpf_printk("not a dns packet...");
```
haesbaert, line 104: filter before doing the filling, this should be way up.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    }
+
+    // constrain the read size to make the verifier happy
+    long readsize = BPF_CORE_READ(skb, len);
+    if (readsize > MAX_DNS_PACKET) {
+        readsize = MAX_DNS_PACKET;
+    }
+
+    // udp_send_skb includes the IP and UDP header, so offset
+    long offset = transport_header_offset + sizeof(struct udphdr);
```
haesbaert, line 115: size_t would fit this better.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    if (ebpf_events_is_trusted_pid())
+        return 0;
+
+    struct ebpf_dns_event *event = bpf_ringbuf_reserve(&ringbuf, sizeof(*event), 0);
+    if (!event)
+        return 0;
+
+    // read from skbuf
+    unsigned char *data = BPF_CORE_READ(skb, head);
+    // get lengths
```
haesbaert, line 62: this is a very confusing name, as the skb itself has a data member.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+
+    // filter out non-dns packets
+    if (event->net.sport != 53 && event->net.dport != 53) {
+        bpf_printk("not a dns packet...");
+        goto out;
+    }
+
+    // constrain the read size to make the verifier happy
+    long readsize = BPF_CORE_READ(skb, len);
+    if (readsize > MAX_DNS_PACKET) {
```
haesbaert, line 110: `len` is not what you want, as what you want is the amount of linear data in the skb, `len` is the total, you want the equivalent of `skb_headlen`, which should be something like:
```
static inline unsigned int skb_headlen(const struct sk_buff *skb)
{
	return skb->len - skb->data_len;
}
```

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+        readsize = MAX_DNS_PACKET;
+    }
+
+    // udp_send_skb includes the IP and UDP header, so offset
+    long offset = transport_header_offset + sizeof(struct udphdr);
+
+    long ret = bpf_probe_read_kernel(event->pkts[0].pkt, readsize, data + offset);
```
haesbaert, line 117: small nit, I'd prefer `data + transport_header_offset + sizeof(udphdr)`, less mental strain remembering what offset is, feel free to ignore.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    if (ret != 0) {
+        bpf_printk("error reading in data buffer: %d", ret);
+        goto out;
+    }
+
+    event->hdr.type = EBPF_EVENT_NETWORK_DNS_PKT;
+    event->udp_evt  = evt_type;
+    bpf_ringbuf_submit(event, 0);
+
```
haesbaert, line 126: as mentioned before, you're writing a fixed size amount of data, instead of taking the size of the packet into account.

GPL/Events/Process/Probe.bpf.c
```diff
@@ -611,8 +606,7 @@ static int tty_write__enter(struct kiocb *iocb, struct iov_iter *from)
     else
         goto out;
 
-    u64 nr_segs = BPF_CORE_READ(from, nr_segs);
-    nr_segs     = nr_segs > MAX_NR_SEGS ? MAX_NR_SEGS : nr_segs;
+    u64 nr_segs = get_iovec_nr_segs_or_max(from);
```
haesbaert, line 610: should be in a different PR.

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -386,6 +386,9 @@ static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj
         err = err ?: bpf_program__set_autoload(obj->progs.kretprobe__vfs_write, false);
         err = err ?: bpf_program__set_autoload(obj->progs.kprobe__chown_common, false);
         err = err ?: bpf_program__set_autoload(obj->progs.kretprobe__chown_common, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kprobe__ip_send_udp, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kprobe__skb_consume_udp, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.kretprobe__skb_consume_udp, false);
```
haesbaert, line 391: kret isn't needed.

GPL/Events/EbpfEventProto.h
```diff
@@ -40,6 +42,9 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
+    EBPF_EVENT_NETWORK_UDP_SEND             = (1 << 20),
+    EBPF_EVENT_NETWORK_UDP_RECV             = (1 << 21),
+    EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 22),
```
haesbaert, line 47: only DNS_PKT is a header type, the other is metadata from DNS_PKT, it shouldn't be on the same space.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    // udp_send_skb includes the IP and UDP header, so offset
+    long offset = transport_header_offset + sizeof(struct udphdr);
+
+    long ret = bpf_probe_read_kernel(event->pkts[0].pkt, readsize, data + offset);
+    if (ret != 0) {
+        bpf_printk("error reading in data buffer: %d", ret);
+        goto out;
+    }
+
```
haesbaert, line 122: as I mentioned before, I'd prefer if we send the whole packet to userland, including ip and udp headers.

GPL/Events/Network/Network.h
```diff
@@ -67,6 +75,9 @@ static int ebpf_sock_info__fill(struct ebpf_net_info *net, struct sock *sk)
     case IPPROTO_TCP:
         net->transport = EBPF_NETWORK_EVENT_TRANSPORT_TCP;
         break;
+    case IPPROTO_UDP:
+        net->transport = EBPF_NETWORK_EVENT_TRANSPORT_UDP;
```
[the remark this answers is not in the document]
haesbaert: This function is also never called for a udp socket, and you fill `ebpf_net_info` incompletely in `handle_consume`

GPL/Events/Network/Probe.bpf.c
haesbaert: `event->net` is incomplete, missing family and transport.

GPL/Events/Network/Probe.bpf.c
haesbaert: can't print things on a per-packet basis.

non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -977,6 +981,10 @@ static void out_net_info(const char *name, struct ebpf_net_event *evt)
         out_string("transport", "TCP");
         out_comma();
         break;
+    case EBPF_NETWORK_EVENT_TRANSPORT_UDP:
+        out_string("transport", "UDP");
+        out_comma();
+        break;
```
haesbaert, line 987: I don't think you ever hit this since net->transport is never filled.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+        return 0;
+    }
+
+    return 0;
+}
+
+SEC("kretprobe/skb_consume_udp")
+int BPF_KRETPROBE(kretprobe__skb_consume_udp, int ret)
+{
+    u64 pid_tid = bpf_get_current_pid_tgid();
```
haesbaert, line 186: don't need kretprobe as mentioned above.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+        return 0;
+    }
+
+    long update_err = bpf_map_update_elem(&pkt_ctx, &pid_tid, &kctx, BPF_ANY);
+    if (update_err != 0) {
+        bpf_printk("error updating context map in udp_recvmsg: %d", update_err);
+        return 0;
+    }
+
+    return 0;
```
haesbaert, line 180: all of this can go away, don't need the context.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -36,24 +36,181 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+        return 0;
+    }
+    return handle_consume(skb, len, EBPF_EVENT_NETWORK_UDP_RECV);
+}
+
+SEC("kprobe/ip_send_skb")
+int BPF_KPROBE(kprobe__ip_send_udp, struct net *net, struct sk_buff *skb)
+{
+    long len = BPF_CORE_READ(skb, len);
+    return handle_consume(skb, len, EBPF_EVENT_NETWORK_UDP_SEND);
```
haesbaert, line 155: everything should be derivable from skb, no need for socket or an additional len.

haesbaert (changes requested):
This needs some work, but it's a step in the right direction imho.
It would be better to clean up the unused bits before asking for a review.
As we discussed yesterday, it's likely you want to use the varlen thingy to express the packet instead of hard coding it.
Let me know if anything is not clear.

GPL/Events/Network/Probe.bpf.c
fearful-symmetry: You mean as opposed to parsing it in ebpf, or in addition to parsing it in ebpf?

GPL/Events/EbpfEventProto.h
```diff
@@ -40,6 +46,7 @@ enum ebpf_event_type {
     EBPF_EVENT_PROCESS_SHMGET               = (1 << 17),
     EBPF_EVENT_PROCESS_PTRACE               = (1 << 18),
     EBPF_EVENT_PROCESS_LOAD_MODULE          = (1 << 19),
+    EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 22),
```
nicholasberlin, line 49:
```suggestion
    EBPF_EVENT_NETWORK_DNS_PKT              = (1 << 20),
```

GPL/Events/Network/Network.h
```diff
@@ -14,6 +14,8 @@
 #define AF_INET 2
 #define AF_INET6 10
 
+#define MSG_PEEK 2
```
nicholasberlin, line 17: I don't see this being used. Delete?

GPL/Events/EbpfEventProto.h
```diff
@@ -379,6 +393,22 @@ struct ebpf_net_event {
     char comm[TASK_COMM_LEN];
 } __attribute__((packed));
 
+struct dns_body {
```
nicholasberlin, line 396: I think this struct can be deleted.

GPL/Events/EbpfEventProto.h
```diff
@@ -379,6 +393,22 @@ struct ebpf_net_event {
...
+
+struct ebpf_dns_event {
+    struct ebpf_event_header hdr;
+    struct ebpf_pid_info pids;
+    struct ebpf_net_info net;
+    char comm[TASK_COMM_LEN];
+    enum ebpf_net_udp_info udp_evt;
+    // Variable length fields: dns body
+    struct ebpf_varlen_fields_start vl_fields;
+    // uint8_t pkt[MAX_DNS_PACKET];
```
nicholasberlin, line 409:
```suggestion
```

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,123 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+}
+
+SEC("fentry/ip_send_skb")
+int BPF_PROG(fentry__ip_send_skb, struct net *net, struct sk_buff *skb)
+{
+    return udp_skb_handle(skb, EBPF_NETWORK_EVENT_IP_SEND_UDP);
+}
+
+SEC("fentry/skb_consume_udp")
+int BPF_PROG(fexit__skb_consume_udp, struct sock *sk, struct sk_buff *skb, int len)
```
nicholasberlin, line 143:
```suggestion
int BPF_PROG(fentry__skb_consume_udp, struct sock *sk, struct sk_buff *skb, int len)
```

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,123 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+}
+
+SEC("kprobe/ip_send_skb")
+int BPF_KPROBE(kprobe__ip_send_udp, struct net *net, struct sk_buff *skb)
+{
+    return udp_skb_handle(skb, EBPF_NETWORK_EVENT_IP_SEND_UDP);
+}
+
+SEC("kprobe/skb_consume_udp")
+int BPF_KPROBE(kprobe__skb_consume_udp, struct net *net, struct sk_buff *skb)
```
nicholasberlin, line 159: Why isn't `len` a parameter here, like the fentry version?

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -403,6 +405,8 @@ static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj
         err = err ?: bpf_program__set_autoload(obj->progs.fexit__do_truncate, false);
         err = err ?: bpf_program__set_autoload(obj->progs.fexit__vfs_write, false);
         err = err ?: bpf_program__set_autoload(obj->progs.fexit__chown_common, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.fentry__ip_send_skb, false);
+        err = err ?: bpf_program__set_autoload(obj->progs.fexit__skb_consume_udp, false);
```
nicholasberlin, line 409:
```suggestion
        err = err ?: bpf_program__set_autoload(obj->progs.fentry__skb_consume_udp, false);
```

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,123 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+        goto out;
+    }
+    ebpf_vl_field__set_size(&event->vl_fields, field, headlen);
+
+    event->hdr.type = EBPF_EVENT_NETWORK_DNS_PKT;
+    event->udp_evt  = evt_type;
+    ebpf_ringbuf_write(&ringbuf, event, EVENT_SIZE(event), 0);
+
+    return 0;
+
```
nicholasberlin, line 131:
```suggestion
```
don't need these lines, since we aren't doing anything in the `out` label.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,123 @@ static int inet_csk_accept__exit(struct sock *sk)
     return 0;
 }
 
+static int udp_skb_handle(struct sk_buff *skb, enum ebpf_net_udp_info evt_type)
+{
+
+    if (ebpf_events_is_trusted_pid())
+        return 0;
```
nicholasberlin, line 51: Returning zero here implies to me that we want to avoid the handling in `out`, but `out` isn't doing anything. Should `goto out` for consistency.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,123 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+
+    struct ebpf_dns_event *event = get_event_buffer();
+    if (!event)
+        return 0;
```
nicholasberlin, line 55: Should `goto out` for consistency.

fearful-symmetry:
Alright, tested on 6.10 and 5.15...

GPL/Events/EbpfEventProto.h
nicholasberlin: This was resolved but I don't see the change. Forgot to commit/push?

GPL/Events/EbpfEventProto.h
nicholasberlin: I don't see this change. Was I wrong? Is it needed?

GPL/Events/EbpfEventProto.h
nicholasberlin: I don't see this change. Why preserve it?

GPL/Events/Network/Probe.bpf.c
nicholasberlin: Should use also use the `len` parameter like the `fentry` version, right?

nicholasberlin:
I think we have settled on the skb functions, which means the previously excellent description needs some love.

GPL/Events/EbpfEventProto.h
fearful-symmetry: Yup, accidentally editing the files in `artifacts/`

GPL/Events/Network/Probe.bpf.c
fearful-symmetry: Ah, you mean check the `len`, thought you meant the parameter was missing from the probe.

nicholasberlin: Oops, should've been more clear. Change LGTM.

haesbaert:
I'll have a look on this tonight

GPL/Events/EbpfEventProto.h
```diff
@@ -11,6 +11,12 @@
 #define EBPF_EVENTPROBE_EBPFEVENTPROTO_H
 
 #define TASK_COMM_LEN 16
+// The theoretical max size of DNS packets over UDP
+// Like so many things in DNS this number probaby isn't 100% accurate.
+// DNS extensions in RFC2671 and RFC6891 mean the actual size can be larger,
+// although an additonal body over 512 is probably due to additional RR fields,
+// which we can (probably) get away with throwing out for now.
+#define MAX_DNS_PACKET 512
```
haesbaert, line 20: 1500 is a better value, in essence a real valyue would be 1500 - sizeof(iphdr) - sizeof(udphdr), but 1500 is round.

fearful-symmetry:
So, did a bit more testing with large DNS queries using this: 
```
dig +noall +answer +comments @166.84.7.99 512.size.dns.netmeister.org a
```

Haven't noticed any weird paging behavior.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,136 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    size_t datalen  = BPF_CORE_READ(skb, data_len);
+    size_t headlen  = readsize - datalen;
+    // headlen of zero indicates we have no non-paged data, and thus cannot read
+    // anything from the root data node
+    if (headlen == 0) {
+        u32 zero                    = 0;
+        struct ebpf_event_stats *es = bpf_map_lookup_elem(&ringbuf_stats, &zero);
+        if (es != NULL) {
+            es->lost++;
+        }
```
haesbaert, line 117: `lost` expresses lost events from kernel -> ringbuffer, feel free to add another member in `ebpf_event_stats`, like `zero_dns` or something.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,136 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    if (headlen > MAX_DNS_PACKET) {
+        headlen = MAX_DNS_PACKET;
+    }
+
+    ebpf_vl_fields__init(&event->vl_fields);
+    struct ebpf_varlen_field *field;
+    field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_DNS_BODY);
+
+    long ret = bpf_probe_read_kernel(field->data, headlen,
+                                     skb_head + transport_header_offset + sizeof(struct udphdr));
```
haesbaert, line 130: not to bikeshed, but didn't we figure this is the same as skb->data?

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,136 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    if (headlen == 0) {
+        u32 zero                    = 0;
+        struct ebpf_event_stats *es = bpf_map_lookup_elem(&ringbuf_stats, &zero);
+        if (es != NULL) {
+            es->lost++;
+        }
+        goto out;
+    }
+
+    if (headlen > MAX_DNS_PACKET) {
```
haesbaert, line 121:
```suggestion
    event->orig_len = headlen;
    if (headlen > MAX_DNS_PACKET) {
```

Just so that userland can detect truncation.

non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -1051,6 +1058,40 @@ static void out_network_connection_accepted_event(struct ebpf_net_event *evt)
...
+
+    printf("\"data\":");
+    out_array_start();
+    struct ebpf_varlen_field *field;
+    FOR_EACH_VARLEN_FIELD(event->vl_fields, field)
+    {
+        printf("packet %d: ", event->udp_evt);
+        for (size_t i = 0; i < field->size; i++) {
+            uint8_t part = field->data[i];
+            printf("%02X ", part);
```
haesbaert, line 1083:
```suggestion
            printf("%02X", part);
```

GPL/Events/EbpfEventProto.h
```diff
@@ -341,13 +347,19 @@ struct ebpf_process_load_module_event {
...
 };
 
 enum ebpf_net_info_af {
     EBPF_NETWORK_EVENT_AF_INET  = 1,
     EBPF_NETWORK_EVENT_AF_INET6 = 2,
 };
 
+enum ebpf_net_udp_info {
+    EBPF_NETWORK_EVENT_SKB_CONSUME_UDP = 1,
+    EBPF_NETWORK_EVENT_IP_SEND_UDP     = 2,
```
haesbaert, line 360: you sure you don't want to rename these to something more generic? like send/recv?

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,136 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+
+        event->net.family = EBPF_NETWORK_EVENT_AF_INET;
+    } else if (ip_hdr.version == 6) {
+        struct ipv6hdr ip6_hdr;
+        bpf_core_read(&ip6_hdr, sizeof(struct ipv6hdr), skb_head + net_header_offset);
+        proto = ip6_hdr.nexthdr;
+
+        bpf_probe_read(event->net.saddr6, 16, ip6_hdr.saddr.in6_u.u6_addr8);
+        bpf_probe_read(event->net.daddr6, 16, ip6_hdr.daddr.in6_u.u6_addr8);
+
```
haesbaert, line 81:
```suggestion

        if (bpf_probe_read(event->net.saddr, 4, &ip_hdr.saddr) != 0) {
            goto out;
        }
        if (bpf_probe_read(event->net.daddr, 4, &ip_hdr.daddr) != 0) {
            goto out;
        }

        event->net.family = EBPF_NETWORK_EVENT_AF_INET;
    } else if (ip_hdr.version == 6) {
        struct ipv6hdr ip6_hdr;
        if (bpf_core_read(&ip6_hdr, sizeof(struct ipv6hdr), skb_head + net_header_offset) != 0) {
            goto out;
        }
        proto = ip6_hdr.nexthdr;

        if (bpf_probe_read(event->net.saddr6, 16, ip6_hdr.saddr.in6_u.u6_addr8) != 0) {
            goto out;
        }
        if (bpf_probe_read(event->net.daddr6, 16, ip6_hdr.daddr.in6_u.u6_addr8) != 0) {
            goto out;
        }

```

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,136 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+        event->net.family = EBPF_NETWORK_EVENT_AF_INET6;
+    }
+
+    if (proto != IPPROTO_UDP) {
+        goto out;
+    }
+
+    struct udphdr udp_hdr;
+    bpf_core_read(&udp_hdr, sizeof(struct udphdr), skb_head + transport_header_offset);
```
haesbaert, line 90:
```suggestion
    if (bpf_core_read(&udp_hdr, sizeof(struct udphdr), skb_head + transport_header_offset) != 0)
        goto out;
```

GPL/Events/Network/Probe.bpf.c
```diff
@@ -43,6 +44,136 @@ static int inet_csk_accept__exit(struct sock *sk)
...
+    event->net.dport     = bpf_ntohs(udp_hdr.dest);
+    event->net.sport     = bpf_ntohs(udp_hdr.source);
+    event->net.transport = EBPF_NETWORK_EVENT_TRANSPORT_UDP;
+
+    // filter out non-dns packets
+    if (event->net.sport != 53 && event->net.dport != 53) {
+        goto out;
+    }
+
```
haesbaert, line 99:
```suggestion

```

Moved the comparison before, this way all the `out` clauses are before we start writing to the ringbuffer.

GPL/Events/Network/Probe.bpf.c
haesbaert: just checking for the return value, rest of the code seems to be consistent.

haesbaert (changes requested):
I didn't look much on the ipv6.
Summary is, the counter should be a new member, as `lost` is used basically to detect when we're too slow to read the ringbuffer.
Consistently check the errors of `bpf_*`, and validate data before mutating state (checking port/src before writign to event)

Maybe @nicholasberlin has more input

GPL/Events/Network/Probe.bpf.c
fearful-symmetry: On some of the methods, no, `head` will start at the IP header, not the application layer body.

GPL/Events/EbpfEventProto.h
fearful-symmetry: Feels a little more descriptive this way, so userland has some data on what we're probing.

GPL/Events/Network/Probe.bpf.c
fearful-symmetry: The github UI doesn't make it too clear where you moved it to, but I took a guess...

GPL/Events/Network/Probe.bpf.c
haesbaert: If it starts at the IP header, how will userland know what to parse?

haesbaert: Wait, maybe I missed something, shouldn't skb->data always be the payload of udp (the dns body)?

testing/test_bins/udp_send.c
```diff
@@ -0,0 +1,50 @@
...
+        buffer[i] = 0xff;
+    }
+}
+
+int main(int argc, char **argv)
+{
+    int sockfd;
+    struct sockaddr_in server;
+
+    create_buffer(buffer, sizeof(buffer));
```
nicholasberlin, line 30: this function could be replaced with:
```
memset(buffer, 0xff, sizeof(buffer));
```

but whatever.

GPL/Events/Network/Probe.bpf.c
fearful-symmetry: Sorry, should have been more clear. So, for `ip_send_skb`, `body` seems to start at the IP header, not the application body. However, the offsets at `skb->network_header` and `skb->transport_header` are still correct, so as long as we read from `skb->head + offsets`, we'll start reading from the beginning of the application body, so it won't start at the IP header.

However, as I type this, I realize it should be possible to just read from `skb->head + skb->transport_header`, which is as bit simpler. Let me try that...

testing/test_bins/udp_send.c
```diff
@@ -0,0 +1,50 @@
...
+#define BUFFER_SIZE 512
+
+char buffer[BUFFER_SIZE];
+
+void create_buffer(uint8_t *buffer, size_t length)
+{
+    for (size_t i = 0; i < length; i++) {
+        buffer[i] = 0xff;
+    }
+}
```
nicholasberlin, line 23:
```suggestion
```
Since you are using `memset`, can delete this.

GPL/Events/Network/Probe.bpf.c
haesbaert: What is body?
Do you mean that skb->data is the start of ip?
Can you just show print skb->data on receive and incoming so we can see?

fearful-symmetry: Nevermind, I confused myself, `skb->head + skb->transport_header` is the _beginning_ of the UDP header, hence `+ sizeof(struct udphdr)`

fearful-symmetry: So, this is the first couple bytes of `data` from `ip_send_skb`:
```
[69, 0, 0, 0, 25, 238, 0, 0, 64, 17, 0, 0, 127, 0, 0, 1, 127, 0, 0, 53, 202, 196, 0, 53, 0, 59, 254, 130, 236, 132, 1, 32, 0, 1, 0, 0, 0, 0, 0, 1, 6, 103, 105, 116, 104, 117, 98, 3, 99, 111, 109]}
```

That looks like the IP header, you can even see the addresses ( `127, 0, 0, 1, 127, 0, 0, 53,`).

This is `data` in `skb_consume_udp`:

```
[236, 132, 1, 32, 0, 1, 0, 0, 0, 0, 0, 1, 6, 103, 105, 116, 104, 117, 98, 3, 99, 111, 109, 0, 0, 1, 0, 1, 0, 0, 41, 4, 208, 0, 0, 0, 0, 0, 12, 0, 10, 0, 8, 82, 206, 183, 136, 107, 47, 147, 50]
```

That starts at the DNS packet; the first two bytes are the query id (236, 132), which you can _also_ see in the first packet, after the headers.

haesbaert (changes requested):
Missing an error case in sendto, and just make the test more idiomatic.
I wonder, why not use the udpsend I wrote in the PR? that way you can test both connected and unconnected, not that it makes a difference now.
At any rate I'm fine with this given this changes.

haesbaert:
I haven't tested the last iteration but looks good to me, good work!

haesbaert (approved):

nicholasberlin (approved):

PR #208, opened 2024-09-20
Refactor ebpf test runner

fearful-symmetry:
## How to run tests

This completely refactors the go testrunner for EventsTrace. This has a number of advantages:

- tests can now be run locally, without bluebox
- You can run individual tests
- Tests are structured in the "standard" golang way, and take advantage of the go stdlib test functionality
- I added a `make testbins` target for the c test binaries

This also fixes a few known bugs, and probably many more:

- The test runner would sometimes eat stderr messages that occur during startup
- If the test runner got overwhelmed with data from stdout/stderr, it would drop the log data
- If a test failed, it would sometimes hang and time out instead of immediately failing

In addition, I fixed a few bugs in the tests that I ran into while refactoring:

- The tty output would try to use the escape character `\x`, which is not a valid JSON escape character and broke all the JSON parsers I tried.
- the overlayfs test would try to remove a directory without unmounting it, which would fail

## How to run tests

```bash
make testbins #make the test binaries in testing/test_bins
cd testing/testrunner
go test -c #compile the test in advance, so we don't need root to have a go environment
# run all tests
sudo ./testrunner.test
# run a single test
sudo ./testrunner.test -test.run TestEbpf/Tcpv6ConnectionClose -test.v
```

## Known issues with individual tests:

- `fork_exec.c` works by setting the capabilities mask to an identifiable but meaningless mask value; for whatever reason this works in the test environment, but on my machine the new capability set completely breaks the binary.
- `create_rename_delete_file_container.c` breaks on newer kernels if you try to remove the directories before doing an unmount on `ovl_mountpoint`, but on the mainline tests, and only the mainline tests (which are actually kind of old), this unmount operation fails with an Invalid Argument error.


testing/testrunner/utils.go
```diff
@@ -180,92 +181,101 @@ type NetConnCloseEvent struct {
...
+	switch runtime.GOARCH {
+	case "amd64":
+		arch = "x86_64"
+	case "arm64":
+		arch = "aarch64"
+	default:
+		fmt.Printf("unsupported arch %s, reverting to root path for test binaries\n", runtime.GOARCH)
+		return
+	}
+	rootEbpfPath := strings.TrimSpace(string(res))
```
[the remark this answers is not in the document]
nicholasberlin: Nevermind, I had it backwards. It will always be set here because of line 196.

testing/testrunner/tests.go
```diff
@@ -1,741 +0,0 @@
...
-	AssertStringsEqual(ev.Net.Family, "AF_INET6")
-	AssertStringsEqual(ev.Net.SourceAddr, "::1")
-	AssertInt64Equal(ev.Net.SourcePort, binOutput.ClientPort)
-	AssertStringsEqual(ev.Net.DestAddr, "::1")
-	AssertInt64Equal(ev.Net.DestPort, binOutput.ServerPort)
-	AssertInt64Equal(ev.Net.NetNs, binOutput.NetNs)
-	AssertStringsEqual(ev.Comm, "tcpv6_connect")
-}
-
-func TestTcFilter() {
```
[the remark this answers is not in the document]
fearful-symmetry: ah, I completely nuked that test. Good catch.

testing/testrunner/tests.go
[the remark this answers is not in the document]
fearful-symmetry: Alright, I tried about 5 different ways of restoring/commiting the files, and git refuses to pick it up as a rename relative to the `main` branch. I think the file is just different enough that it's a "new" file. Go wants a test suite to have the path `*_test.go`, so it needs to be renamed to something.

nicholasberlin (approved):

Makefile
```diff
@@ -170,5 +173,14 @@ endif
 	go install github.com/florianl/bluebox@b8590fb1850f56df6e6d7786931fcabdc1e9173d
 	cd testing && ./run_tests.sh ${ARCH} ${ARTIFACTS_PATH} ${PWD}/kernel-images/${IMG_FILTER}/${ARCH}/*
 
+testbins: testbinpath $(TESTBIN_PROGS)
+
+testbinpath: 
+	mkdir -p testing/test_bins/bin/${ARCH}
```
haesbaert, line 179:
```suggestion
testbinpath:
	mkdir -p testing/test_bins/bin/${ARCH}
```

non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -280,7 +280,10 @@ static void out_escaped_string(const char *value)
             break;
         default:
             if (!isascii(c) || iscntrl(c))
-                printf("\\x%02x", c);
+                // \x is not a valid escape character in json,
+                // and something like '\xff' will break a remarkable number of JSON parsers.
+                // we have to print as '0xff'
+                printf("0x%02x", (uint8_t)c);
             else
```
haesbaert, line 287:
```suggestion
                printf("0x%02x", c);
            else
```

testing/README.md
```diff
@@ -55,6 +55,22 @@ By default `run_tests.sh` will pass `-j$(nproc)` to `parallel` (i.e. spin up as
...
+
+```
+#compile the test in advance, so we don't need root to have a go environment
+go test -c
+# run all tests
+sudo ./testrunner.test
+# run a single test
+sudo ./testrunner.test -test.run TestEbpf/Tcpv6ConnectionClose
+# run in verbose mode 
+sudo ./testrunner.test -test.run TestEbpf/Tcpv6ConnectionClose -test.v
```
haesbaert, line 71:
```suggestion
sudo ./testrunner.test -test.run TestEbpf/Tcpv6ConnectionClose
# run in verbose mode
sudo ./testrunner.test -test.run TestEbpf/Tcpv6ConnectionClose -test.v
```

haesbaert (approved):
I made three small suggestions, basically whitespace.
I was able to run the tests, but it took me awhile to figure why it wasn't working initially, I had a /tmp/foo already in place.

non-GPL/Events/EventsTrace/EventsTrace.c
fearful-symmetry: Are you suggesting I remove the cast? I deliberately added that, since usually when I try to hex-print a char value in C, I run into this: https://stackoverflow.com/questions/45776088/why-printf-show-false-value-of-an-hex-number

nicholasberlin: Doesn't the `%02` takes care of that?

haesbaert: Oh my bad, I didn't notice that they were using a char for `c` which gets signed extended,
`%02` only means "pad with zero and print at least 2 characters".
There's two ways to fix it
 1 - remove the cast and change the `char c` to `int c`, that's the most normal thing people do, in libc all "getchar()" like things store it in an int.
 2 - print it with `printf("0x%02hhx", c)`

haesbaert: or just leave as it is, disregard my initial suggestion.

PR #209, opened 2024-10-22
Use the proper struct for syscall tracepoint probes

nicholasberlin:
See https://github.com/inspektor-gadget/inspektor-gadget/issues/2444#issuecomment-1966440483
and see https://github.com/inspektor-gadget/inspektor-gadget/pull/2546
and see https://lore.kernel.org/lkml/20231005123413.GA488417@alecto.usersys.redhat.com/t/#u

This change fixes test failures in RHEL 9.


fearful-symmetry (approved):
Interesting bug!

nfritts (approved):

PR #213, opened 2024-11-19
Include namespace inode numbers on process events

haesbaert:
This diff adds the inode number of each namespace the process belongs to, to process events. Two namespaces were excluded: pid_for_children and time_for_children, see namespaces(7).

The pid namespace is a bit special as it's not in nsproxy, the user facing pid level is the last/deepest.

There was an inclusion of mnt ns, that is not used by beats or endpoint, I kept it but made it fetch from the new "path".

While here, add support for changing the compiler in gen_initramfs.sh by setting CC. While Fedora includes a ${arch}-linux-gnu-gcc, it doesn't include matching headers, so I can't really compile the tests locally.

### DIFF
--- a/GPL/Events/EbpfEventProto.h
+++ b/GPL/Events/EbpfEventProto.h
@@ -157,6 +157,16 @@ struct ebpf_file_info {
     uint64_t ctime;
 } __attribute__((packed));
 
+struct ebpf_namespace_info {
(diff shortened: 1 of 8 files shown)


nicholasberlin (approved):
LGTM

stanek-michal (approved):
LGTM

PR #219, opened 2025-02-06
Fix memfd probes offset. Fixes #214

### LINKED ISSUE #214: tracepoint/syscalls/sys_enter_memfd_create and tracepoint/syscalls/sys_enter_shmget assume incorrect format
### Severity

Major

### Affects Version(s)

8.5.0, 8.5.1, 8.6.0

### Epic/Meta Issue

_No response_

### Kernel Version(s)

lala

### Linux Distro

RHEL9

### Fix Versions

_No response_

### Description

Code incorrectly assumes a fixed format for the two tracepoints [here](https://github.com/elastic/ebpf/blob/main/GPL/Events/Process/Probe.bpf.c#L423) and [here](https://github.com/elastic/ebpf/blob/main/GPL/Events/Process/Probe.bpf.c#L382). RHEL9 stashes a new member that changes alignment

    But here is how it looks like on RHEL9 where the common area has an extra member:

            field:unsigned short common_type;       offset:0;       size:2; signed:0;
            field:unsigned char common_flags;       offset:2;       size:1; signed:0;
            field:unsigned char common_preempt_count;       offset:3;       size:1; signed:0;
            field:int common_pid;   offset:4;       size:4; signed:1;
    --->    field:unsigned char common_preempt_lazy_count;  offset:8;       size:1; signed:0;

            field:__data_loc char[] filename;       offset:12;      size:4; signed:1;
            field:pid_t pid;        offset:16;      size:4; signed:1;
            field:pid_t old_pid;    offset:20;      size:4; signed:1;

Fix is to fetch the offsets via CORE like the rest of the code does.

haesbaert:
We should get args from BTF, not hardcode them, RHEL9 for instance stashes a `common_preempt_lazy_count` which pushes the probe specific data.

```
        field:unsigned short common_type;       offset:0;       size:2; signed:0;
        field:unsigned char common_flags;       offset:2;       size:1; signed:0;
        field:unsigned char common_preempt_count;       offset:3;       size:1; signed:0;
        field:int common_pid;   offset:4;       size:4; signed:1;
--->    field:unsigned char common_preempt_lazy_count;  offset:8;       size:1; signed:0;
        field:__data_loc char[] filename;       offset:12;      size:4; signed:1;
        field:pid_t pid;        offset:16;      size:4; signed:1;
        field:pid_t old_pid;    offset:20;      size:4; signed:1;
```

### DIFF
--- a/GPL/Events/Process/Probe.bpf.c
+++ b/GPL/Events/Process/Probe.bpf.c
@@ -390,22 +390,18 @@ out:
 }
 
 SEC("tracepoint/syscalls/sys_enter_shmget")
-int tracepoint_syscalls_sys_enter_shmget(struct syscall_trace_enter *ctx)
+int tracepoint_syscalls_sys_enter_shmget(struct trace_event_raw_sys_enter *ctx)
 {
     if (ebpf_events_is_trusted_pid())
         goto out;
 
     struct shmget_args {
-        short common_type;
-        char common_flags;
-        char common_preempt_count;
-        int common_pid;
-        int __syscall_nr;
         long key;
         size_t size;
         long shmflg;
     };
-    struct shmget_args *ex_args    = (struct shmget_args *)ctx;
+
+    struct shmget_args *ex_args    = (struct shmget_args *)BPF_CORE_READ(ctx, args);
     const struct task_struct *task = (struct task_struct *)bpf_get_current_task();
 
     if (is_kernel_thread(task))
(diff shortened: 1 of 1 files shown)


nfritts (approved):

GPL/Events/Process/Probe.bpf.c
```diff
@@ -390,22 +390,18 @@ int BPF_KPROBE(kprobe__ptrace_attach,
 }
 
 SEC("tracepoint/syscalls/sys_enter_shmget")
-int tracepoint_syscalls_sys_enter_shmget(struct syscall_trace_enter *ctx)
+int tracepoint_syscalls_sys_enter_shmget(struct trace_event_raw_sys_enter *ctx)
```
nicholasberlin, line 393: Have you seen this? https://github.com/inspektor-gadget/inspektor-gadget/issues/2444#issuecomment-1966440483

(From this PR: https://github.com/elastic/ebpf/pull/209)

haesbaert: Ack, I'm withdrawing this PR after our meeting yesterday.

I hadn't realized there were two different structures, and using CORE on the ctx's zero len array is not trivial, we need to get btf_size_type and do an ugly dance.

```
*ON* RHEL9
struct trace_entry {
	short unsigned int type;
	unsigned char flags;
	unsigned char preempt_count;
	int pid;
	char common_preempt;
};

struct syscall_trace_enter {
	struct trace_entry ent;        [0  - 12[
	int nr;                        [12 - 16[
	long unsigned int args[0];     [16 - ..[
};

struct trace_event_raw_sys_enter {
	struct trace_entry ent;        [0  - 12[
	PAD0                           [12 - 16[
	long int id;                   [16 - 24[
	long unsigned int args[6];     [24 - (24 + 6 * 8)[
	char __data[0];                [32 - ..[
};

*NOT* RHEL9
struct syscall_trace_enter {
	struct trace_entry ent;        [0  -  8[
	int nr;                        [8  - 12[
	PAD0                           [12 - 16[ 
	long unsigned int args[0];     [16 - 16[
};

struct trace_event_raw_sys_enter {
	struct trace_entry ent;        [0  -  8[
	long int id;                   [8  - 16[
	long unsigned int args[6];     [16 - ..[
	char __data[0];                [.. - ..[
};

```

You can see that the `args` offset on both structures on both RHEL9 and not RHEL9 ends up at 16.
It's all a bit of sheer luck since since only one structure gets padded, so the `trace_enter` version still works. It's still a bit "wrong" since we use hard coded offsets.

PR #220, opened 2025-02-10
Fix TCP close events for "idle" connections. Fixes #217

### LINKED ISSUE #217: CLOSE@TCP not issued if there was no traffic
### Severity

Minor

### Affects Version(s)

8.5.0, 8.5.1, 8.6.0

### Epic/Meta Issue

_No response_

### Kernel Version(s)

lala

### Linux Distro

all

### Fix Versions

_No response_

### Description

We don't send the closing event if there was no traffic, this is intentional but it's bad behaviour.
The user likely allocated something when it got a connect/accept, but it is given indication of the end of life cycle.

haesbaert:
Change the condition to supress a close, before it would supress any sockets that had no traffic.

Now we keep state of sockets we sent an attempt/accept and supress the ones we didn't track _AND_ had no traffic. The map is limited, once we have 128k active sockets we can't track them, so fallback to looking at bytes_received and bytes_sent.

We don't really look at tgid, but we will on upcoming probes. Maps are not small, most of the allocation size goes to the hash bucket. The empty map wires 2MB of memory.

6306: hash  name sk_to_tgid  flags 0x1
        key 8B  value 4B  max_entries 131072  memlock 2098760B
        btf_id 5117
        pids EventsTrace(2868637)

### DIFF
--- a/GPL/Events/Network/Probe.bpf.c
+++ b/GPL/Events/Network/Probe.bpf.c
@@ -21,6 +21,14 @@
 
 DECL_FUNC_RET(inet_csk_accept);
 
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, 131072);
+    __type(key, struct sock *);
+    __type(value, u32);
+    __uint(map_flags, BPF_F_NO_PREALLOC);
+} sk_to_tgid SEC(".maps");
+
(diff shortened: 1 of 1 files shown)


nicholasberlin (approved):
LGTM

stanek-michal (approved):

PR #222, opened 2025-02-17
New dns - draft

haesbaert:
just pushing to use the CI, as the local test goes silent.

### DIFF
--- a/GPL/Events/Network/Probe.bpf.c
+++ b/GPL/Events/Network/Probe.bpf.c
@@ -469,7 +469,7 @@ int skb_in_or_egress(struct __sk_buff *skb, int ingress)
 
 	ebpf_vl_fields__init(&event->vl_fields);
 	field = ebpf_vl_field__add(&event->vl_fields, EBPF_VL_FIELD_DNS_BODY);
-	if (bpf_skb_load_bytes(skb, 0, field->data, cap_len))
+	if (bpf_skb_load_bytes(skb, 0, field->data, MAX_DNS_PACKET))
 		goto ignore;
 	ebpf_vl_field__set_size(&event->vl_fields, field, cap_len);
 
--- a/GPL/Events/Network/Probe.bpf.c
+++ b/GPL/Events/Network/Probe.bpf.c
@@ -357,3 +357,182 @@ int BPF_KPROBE(kprobe__tcp_close, struct sock *sk, long timeout)
 {
     return tcp_close__enter(sk);
 }
+
+/* XXX naive, only handles ROUTING and DEST, untested */
+int skb_peel_nexthdr(struct __sk_buff *skb, u8 wanted)
+{
+	struct ipv6hdr ip6;
+	int off;
+	u16 next;
+
+	off = 0;
+	if (bpf_skb_load_bytes(skb, off, &ip6, sizeof(ip6)))
+		return (-1);
+	off += sizeof(ip6);
+	next = ip6.nexthdr;
+
+	for (;;) {
+		if (next == wanted)
+			return (off);
+		switch (next) {
+		case NEXTHDR_ROUTING: /* FALLTHROUGH */
+		case NEXTHDR_DEST:
+			if (bpf_skb_load_bytes(skb, off, &next, sizeof(next)))
+				return (-1);
+			off += (next >> 8) + 1;
+			next = next & 0xff;
+		default:
+			return (-1);
+		}
+	}
+}
+
+int skb_in_or_egress(struct __sk_buff *skb, int ingress)
+{
+	struct udphdr udp;
+	struct bpf_sock *sk, *sk2;
+	u32 *tgid, cap_len;
+	struct ebpf_dns_event2 *event;
+	struct ebpf_varlen_field *field;
+
+	if (skb->family != AF_INET && skb->family != AF_INET6)
+		goto ignore;
+	if ((sk = skb->sk) == NULL)
+		goto ignore;
+	if ((sk = bpf_sk_fullsock(sk)) == NULL)
+		goto ignore;
+	if (sk->family != AF_INET && sk->family != AF_INET6)
+		goto ignore;
+	if (sk->protocol != IPPROTO_UDP)
+		goto ignore;
+
+	if (sk->family == AF_INET) {
+		struct iphdr ip;
+
+		if (bpf_skb_load_bytes(skb, 0, &ip, sizeof(ip))) {
+			bpf_printk("copy error 1");
+			goto ignore;
+		}
+		if (bpf_skb_load_bytes(skb, ip.ihl << 2, &udp, sizeof(udp))) {
+			bpf_printk("copy error 2");
+			goto ignore;
+		}
+		if (ip.protocol != IPPROTO_UDP)
+			goto ignore;
+	} else if (sk->family == AF_INET6) {
+		int t_off;
+
+		t_off = skb_peel_nexthdr(skb, NEXTHDR_UDP);
+		if (t_off == -1)
+			goto ignore;
+
+		if (bpf_skb_load_bytes(skb, t_off, &udp, sizeof(udp))) {
+			bpf_printk("copy error 4");
+			goto ignore;
+		}
+	}
+
+	if (bpf_ntohs(udp.dest) != 53 && bpf_ntohs(udp.source) != 53)
+		goto ignore;
+
+	sk2 = sk;
+	tgid = bpf_map_lookup_elem(&sk_to_tgid, &sk2);
+	if (tgid == NULL) {
+		bpf_printk("udp egress not found");
+		goto ignore;
+	}
+	bpf_printk("%d: udp src=%d dst=%d len=%d",
+	    tgid == NULL ? 0 : *tgid, bpf_ntohs(udp.source),
+	    bpf_ntohs(udp.dest), bpf_ntohs(udp.len));
+
+	cap_len = skb->len;
+	/*
+	 * verifier will complain, even with a skb->len
+	 * check at the beginning.
+	 */
+	if (cap_len == 0)
+		goto ignore;
+	else if (cap_len > MAX_DNS_PACKET)
+		cap_len = MAX_DNS_PACKET;
+
+	event = get_event_buffer();
+	if (event == NULL)
+		goto ignore;
+
+	event->hdr.type = EBPF_EVENT_NETWORK_DNS_PKT;
+	event->hdr.ts = bpf_ktime_get_ns();
+	event->hdr.ts_boot = bpf_ktime_get_boot_ns_helper();
+	event->tgid = *tgid;
+	event->cap_len = cap_len;
+	event->orig_len = skb->len;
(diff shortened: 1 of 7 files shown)


GPL/Events/Network/Probe.bpf.c
```diff
@@ -357,3 +197,230 @@ int BPF_KPROBE(kprobe__tcp_close, struct sock *sk, long timeout)
...
+    if ((sk = bpf_sk_fullsock(sk)) == NULL)
+        goto ignore;
+    if (sk->protocol != IPPROTO_UDP)
+        goto ignore;
+
+    if (sk->family == AF_INET) {
+        struct iphdr ip;
+
+        if (bpf_skb_load_bytes(skb, 0, &ip, sizeof(ip))) {
+            bpf_printk("copy error 1");
```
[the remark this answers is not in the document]
nicholasberlin: or remove

non-GPL/Events/EventsTrace/EventsTrace.c
```diff
@@ -1087,16 +1087,19 @@ static void out_network_connection_accepted_event(struct ebpf_net_event *evt)
...
-    out_pid_info("pids", &event->pids);
+    out_int("tgid", event->tgid);
     out_comma();
 
-    out_net_info("net", &event->net, &event->hdr);
+    out_int("cap_len", event->cap_len);
     out_comma();
 
-    out_string("comm", (const char *)&event->comm);
+    out_int("orig_len", event->cap_len);
```
nicholasberlin, line 1099:
```suggestion
    out_int("orig_len", event->orig_len);
```

testing/testrunner/ebpf_test.go
```diff
@@ -680,31 +680,53 @@ func Tcpv6ConnectionClose(t *testing.T, et *Runner) {
...
 }
 
 func DNSMonitor(t *testing.T, et *Runner) {
-	binOutput := NetBinOut{}
-	runTestUnmarshalOutput(t, "udp_send", &binOutput)
+	addr, err := net.ResolveUDPAddr("udp", "127.0.0.1:53")
+	require.NoError(t, err)
+
+	listen, err := net.ListenUDP("udp", addr)
+	require.NoError(t, err)
```
nicholasberlin, line 690:
```suggestion
	require.NoError(t, err)
	defer listen.Close()
```

testing/testrunner/ebpf_test.go
```diff
@@ -680,31 +680,53 @@ func Tcpv6ConnectionClose(t *testing.T, et *Runner) {
...
+
+	conn, err := net.DialUDP("udp", nil, addr)
+	require.NoError(t, err)
```
nicholasberlin, line 693:
```suggestion
	require.NoError(t, err)
	defer conn.Close()
```

GPL/Events/Network/Probe.bpf.c
```diff
@@ -357,3 +197,230 @@ int BPF_KPROBE(kprobe__tcp_close, struct sock *sk, long timeout)
...
+
+    for (;;) {
+        if (next == wanted)
+            return (off);
+        switch (next) {
+        case NEXTHDR_ROUTING: /* FALLTHROUGH */
+        case NEXTHDR_DEST:
+            if (bpf_skb_load_bytes(skb, off, &next, sizeof(next)))
+                return (-1);
+            off += (next >> 8) + 1;
```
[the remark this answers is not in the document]
haesbaert: marking as resolved as we discussed this yesterday.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -357,3 +197,230 @@ int BPF_KPROBE(kprobe__tcp_close, struct sock *sk, long timeout)
...
+    sk_addr = bpf_map_lookup_elem(&scratch64, &zero);
+    if (sk_addr == NULL)
+        goto ignore;
+    *sk_addr = (u64)sk;
+    tgid     = bpf_map_lookup_elem(&sk_to_tgid, sk_addr);
+    if (tgid == NULL) {
+        bpf_printk("udp egress not found");
+        goto ignore;
+    }
+    bpf_printk("%d: udp src=%d dst=%d len=%d", tgid == NULL ? 0 : *tgid, bpf_ntohs(udp.source),
```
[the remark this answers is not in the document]
haesbaert: coming in the next commit.

GPL/Events/Network/Probe.bpf.c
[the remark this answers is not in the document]
haesbaert: done

GPL/Events/Network/Probe.bpf.c
```diff
@@ -357,3 +197,230 @@ int BPF_KPROBE(kprobe__tcp_close, struct sock *sk, long timeout)
...
+        struct iphdr ip;
+
+        if (bpf_skb_load_bytes(skb, 0, &ip, sizeof(ip))) {
+            bpf_printk("copy error 1");
+            goto ignore;
+        }
+        if (ip.protocol != IPPROTO_UDP)
+            goto ignore;
+        if (bpf_skb_load_bytes(skb, ip.ihl << 2, &udp, sizeof(udp))) {
+            bpf_printk("copy error 2");
```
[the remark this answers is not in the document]
haesbaert: gonna zap all printks

GPL/Events/Network/Probe.bpf.c
```diff
@@ -357,3 +197,230 @@ int BPF_KPROBE(kprobe__tcp_close, struct sock *sk, long timeout)
...
+    /*
+     * Needed for kernels prior to f79efcb0075a20633cbf9b47759f2c0d538f78d8
+     * bpf: Permits pointers on stack for helper calls
+     */
+    sk_addr = bpf_map_lookup_elem(&scratch64, &zero);
+    if (sk_addr == NULL)
+        return (1);
+    *sk_addr = (u64)sk;
+    if (bpf_map_update_elem(&sk_to_tgid, sk_addr, &tgid, BPF_ANY) == 0)
+        bpf_printk("sk %p saved", sk);
```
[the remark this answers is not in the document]
haesbaert: done

non-GPL/Events/EventsTrace/EventsTrace.c
haesbaert: great eyes, many thanks!

testing/testrunner/ebpf_test.go
haesbaert: ooopsy

haesbaert:
I think I've addressed all points, thanks for spotting them!

nicholasberlin (approved):
Changes LGTM. Thanks.

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -405,13 +404,38 @@ static inline int probe_set_autoload(struct btf *btf, struct EventProbe_bpf *obj
...
     }
 
     return err;
 }
 
+static int probe_attach_cgroup(struct EventProbe_bpf *obj)
+{
+    int cgroup_fd;
+
+    cgroup_fd = open("/sys/fs/cgroup", O_RDONLY);
```
fearful-symmetry, line 416: It feels like there's some kind of 1% edge case here with us hard-coding the root path. I think the "safe" way to do this would be to check the filesystem mounts for the cgroups root, which is what beats does.

GPL/Events/Network/Probe.bpf.c
```diff
@@ -357,3 +197,222 @@ int BPF_KPROBE(kprobe__tcp_close, struct sock *sk, long timeout)
...
+        if (bpf_skb_load_bytes(skb, 0, &ip, sizeof(ip)))
+            goto ignore;
+        if (ip.protocol != IPPROTO_UDP)
+            goto ignore;
+        if (bpf_skb_load_bytes(skb, ip.ihl << 2, &udp, sizeof(udp)))
+            goto ignore;
+    } else {
+        goto ignore;
+    }
+#ifdef notyet
```
fearful-symmetry, line 265: Can we get a comment here on why we're ifdef'ing this out?  I assume this is just ipv6 support?

GPL/Events/Network/Probe.bpf.c
```diff
@@ -357,3 +197,222 @@ int BPF_KPROBE(kprobe__tcp_close, struct sock *sk, long timeout)
...
+    sk_addr = bpf_map_lookup_elem(&scratch64, &zero);
+    if (sk_addr == NULL)
+        return (1);
+    *sk_addr = (u64)sk;
+    bpf_map_update_elem(&sk_to_tgid, sk_addr, &tgid, BPF_ANY);
+
+    return (1);
+}
+
+SEC("cgroup/sendmsg4")
```
fearful-symmetry, line 374: Can we put a comment here about how we're also monitoring `{sendmsg4,recvmsg4,connect4}` in order to catch cases where a connection is passed to a child?

fearful-symmetry (commented):
Mostly interesting in the documentation here, but I feel like the cgroup root thing is worth investigating.

non-GPL/Events/Lib/EbpfEvents.c
haesbaert: Are you sure this is what beat does? I couldn't find it where it parses the filesystem, I see that the root can be specified other than the default:
https://github.com/elastic/beats/blob/main/libbeat/processors/add_process_metadata/add_process_metadata.go#L58-L59

It makes sense to have an option to specify a different path, but I don't think we should be parsing mountpoints and searching for a cgroup.
I'd also say that for now, just not supporting those cases is fine.

GPL/Events/Network/Probe.bpf.c
haesbaert: I'd rather expand the commit message, but if you prefer a comment, I can add a comment.

non-GPL/Events/Lib/EbpfEvents.c
fearful-symmetry: This probably isn't a great example, since the beats cgroup monitoring is much more granular and has to worry about containers, but we generally parse mountinfo to get the cgroups: https://github.com/elastic/elastic-agent-system-metrics/blob/main/metric/system/cgroup/util.go#L205

I don't think it's a huge deal here, since if we're not supporting containers or caring about individual cgroups, that removes a massive amount of complexity, but having to deal with years of bizarre cgroup edge cases from customers has made me a tad paranoid.

GPL/Events/Network/Probe.bpf.c
fearful-symmetry: Yeah, that would be fine.

GPL/Events/Network/Probe.bpf.c
fearful-symmetry: Yeah, it's a matter of personal preference, I'm more a "docs in code" person than a "docs in commit" person, but at least a one-liner explaining it might be good.

fearful-symmetry (approved):

PR #223, opened 2025-03-27
Fix socket events for 4.18x kernels

haesbaert:
Related to https://github.com/elastic/quark/issues/135

sk_protocol was a 32bit bitfield in 4.18x, then they moved it to a 16bit word. This fixes it so that it works on both cases, in the old case, the field is considered to be 32bit, in the new 16bit.


haesbaert:
Reproduce it in quark with
```
make initramfs.gz && ./krun.sh initramfs.gz kernel-images/amd64/linux-4.18.0-553.16.1.el8_10.x86_64  quark-test -b t_sock_conn
```

nicholasberlin (approved):

nfritts (approved):

PR #231, opened 2025-09-10
Remove DNS probes and associated code

nicholasberlin:
I had previously thought simply disabling DNS probe loading was enough, but I still see customer errors related to `skb_egress` even after loading is disabled.

This PR simply removes all the DNS probes and related infrastructure.

Elastic Endpoint use [quark](https://github.com/elastic/quark) for DNS probes, so removing them from here.

### DIFF
--- a/non-GPL/Events/Lib/EbpfEvents.c
+++ b/non-GPL/Events/Lib/EbpfEvents.c
@@ -2,7 +2,7 @@
 
 /*
  * Copyright 2021 Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under
-
+ * one or more contributor license agreements. Licensed under the Elastic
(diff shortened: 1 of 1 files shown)


haesbaert (approved):

non-GPL/Events/Lib/EbpfEvents.c
```diff
@@ -2,7 +2,7 @@
 
 /*
  * Copyright 2021 Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under
- * one or more contributor license agreements. Licensed under the Elastic
+
```
stanek-michal, line 5: this needed?

stanek-michal (approved):
ACK (but check https://github.com/elastic/ebpf/pull/231#discussion_r2338043272)

PR #233, opened 2026-03-10
Switch host isolation eBPF probes to use skeleton headers

stanek-michal:
Two main changes:

1) Switch to skeleton headers to be the same as our Events probes
upside: no need to ship per-arch.o objects, no need to rely on fixed path with the objects
2) change TCA_BPF_NAME for TcFilter so it can be easily identified when enumerating tc probes on the host

### DIFF
--- a/non-GPL/HostIsolation/Lib/TcLoader.c
+++ b/non-GPL/HostIsolation/Lib/TcLoader.c
@@ -473,9 +473,9 @@ out:
 
 int netlink_filter_add_end(int fd, struct netlink_ctx *ctx)
 {
-    struct nlmsghdr *nl = NULL;
+    struct nlmsghdr *nl       = NULL;
     struct bpf_prog_info info = {};
-    unsigned int info_len = sizeof(info);
+    unsigned int info_len     = sizeof(info);
     char buf[128];
     int rv  = -1;
     int len = 0;
--- a/non-GPL/HostIsolation/Lib/TcLoader.c
+++ b/non-GPL/HostIsolation/Lib/TcLoader.c
@@ -14,6 +14,7 @@
 
 #include "Common.h"
 #include <argp.h>
+#include <bpf/bpf.h>
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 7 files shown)


non-GPL/HostIsolation/Lib/TcLoader.c
```diff
@@ -470,25 +471,33 @@ int netlink_filter_add_begin(struct netlink_ctx *ctx, const char *ifname)
...
+    if (rv < 0) {
+        ebpf_log("netlink_filter_add_end error: failed to get info by fd\n");
+        goto out;
+    }
+
     nl = &ctx->msg.n;
     memset(buf, 0, sizeof(buf));
 
-    len = snprintf(buf, sizeof(buf), "%s:[.text]", ebpf_obj_filename);
+    len = snprintf(buf, sizeof(buf), "el-endpo_%s:[%u]", info.name, info.id);
```
nicholasberlin, line 498: Just curious, what is `el-endpo_`?

nicholasberlin (approved):
LGTM, there are formatting issues though.


Here's what my bot thinks. None of the suggested changes look necessary to me, as it suggests.

# PR Review: #233 — Switch host isolation eBPF probes to use skeleton headers

**Author:** stanek-michal
**Branch:** pr11704-modern-skeleton-ebpf -> main
**Commits:** 2 | **Files changed:** 7 | **+108 -120**
**URL:** https://github.com/elastic/ebpf/pull/233

## Summary

This PR migrates the Host Isolation eBPF probes (KprobeConnectHook and TcFilter) from runtime `.o` file loading to compile-time skeleton headers, aligning them with the Events probes. It also parameterizes the TcLoaderDemo interface name (previously hardcoded to `ens33`) and changes the `TCA_BPF_NAME` to a structured format (`el-endpo_<name>:[<id>]`) for easier identification when enumerating tc probes. The changes are clean and well-structured.

## Findings

### Significant

1. **Skeleton retry loop does not reset `ctx` between attempts (KprobeConnectHookDemo.c)**

   In `main()`, the retry loop calls `try_load_ebpf_kprobe(load_method, &ctx)` repeatedly. On failure, `try_load_ebpf_kprobe` destroys the local skeleton and does *not* write to `*ctx_out`, so `ctx` in `main` stays at its initialized `NULL` — this is correct. However, the pattern is subtle and relies on an implicit contract that the function never modifies `*ctx_out` on failure. A defensive `*ctx_out = NULL` on the error path inside `try_load_ebpf_kprobe` would make this contract explicit and guard against future regressions.

2. **`bpf_obj_get_info_by_fd` sets `rv = 0` on success path, masking intent (TcLoader.c:489)**

   After `rv = bpf_obj_get_info_by_fd(fd, &info, &info_len)`, `rv` is 0 on success. This is fine functionally, but later at line 518 `rv = 0` is set again redundantly. The dual assignment isn't a bug, but it makes the success flow slightly harder to follow — the original pattern of initializing `rv = -1` and only setting `rv = 0` at the single exit point was cleaner. Consider assigning the `bpf_obj_get_info_by_fd` result to a separate variable (or just checking `< 0` with an inline expression) to preserve the single-exit-point `rv = 0` pattern used elsewhere in this file.

### Minor

3. **`TCA_BPF_NAME` buffer sizing is safe but tight for future changes (TcLoader.c:498)**

   The format `"el-endpo_%s:[%u]"` with `info.name` (max 15 chars from `BPF_OBJ_NAME_LEN`) and `info.id` (max ~10 digits) fits comfortably in the 128-byte buffer (~38 bytes worst case). The truncation check is correctly in place. No action needed — just noting this is well-handled.

4. **Docs update correctly reflects the new CLI (hostisolation.md)**

   The documentation now shows `<iface>` as a required argument and updates the unload command accordingly. Looks good.

## Verdict

**APPROVE** — The migration to skeleton headers is straightforward, the removed `ebpf_open_object_file` and `ebpf_load_and_attach_kprobe` helpers are cleanly replaced by skeleton APIs, and the TcLoaderDemo interface parameterization is a nice usability improvement. The findings above are minor polish suggestions, not blockers.

non-GPL/HostIsolation/Lib/TcLoader.c
nicholasberlin: Oh, I see this now:

> As a bonus, on the eBPF submodule side the TCA_BPF_NAME was changed to add a prefix "el-endpo_" which will help us identify our tc filter when enumerating them on the system. This will be used in another PR to make our qdisc management less conflicting in case user has other tc probes.

nfritts (approved):

PR #234, opened 2026-07-14
Tc filter selective removal

stanek-michal:
Add functions to check if a TC filter exists, and to selectively delete one based on name, without touching other TC filters.


stanek-michal:
Added more robustness fixes in separate commits. Compared with official kernel code for all the APIs as well. Compatibility remains wide as before, 4.13+

stanek-michal:
Fixed remaining issues - DUMP_INTR should now be handled properly everywhere, also if a `qdisc` already exists and is not `clsact` type (the one we need) - we bail. If one exists and is `clsact` we reuse it. If no `qdisc` existed we create one but we never clean it up - safer that way and there shouldn't be any issues holding one, no overhead (cause no filter).

nicholasberlin (approved):
Bots are happy, and LGTH(uman)

fearful-symmetry (approved):

